PR #1, opened 2020-07-28
Implement S3 backend for go-nix

Rizary:
This PR purpose is to add S3 backend for go-nix.


go.sum
```diff
@@ -348,6 +348,63 @@ gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 h1:qIbj1fsPNlZgppZ+VLlY7N33
 gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/errgo.v2 v2.1.0/go.mod h1:hNsd1EY+bozCKY1Ytp96fpM3vjJbqLJn88ws8XvfDNI=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+=======
```
zimbatm, line 351: this looks suspiciously like a merge conflict entry that wasn't removed:
```suggestion
```

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"io"
+	"os"
+	"net/url"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	"log"
+)
+
+// S3BinaryCacheStore ...
```
zimbatm, line 14: Please fix the comment to match the struct name

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+type S3Store struct {
+	Profile string // The name of the AWS configuration profile to use.
```
zimbatm, line 16: comments usually go in front of the attribute

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+	Region string // "aws-region"
```
zimbatm, line 17: Where is the region attribute being used in the code?

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+	Scheme string //  "The scheme to use for S3 requests, https by default."
```
zimbatm, line 18: Same, where is it being used?

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+	// narinfoCompression string // "compression method for .narinfo files"
+	// multipartUpload bool
+	// bufferSize uint64
+ // do we need to put endpoint in here
+ // anything that is needed in minio, no need here. Check.
+	bucketName string
+  client *minio.Client
```
zimbatm, line 25: run `go fmt` to keep the code indented

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// SDK should be in environment
+func (s *S3Store) getUri() string {
+  return "s3://" + s.bucketName
+}
+
+func (s *S3Store) FileExists(ctx context.Context, path string) (bool, error) {
+
+	return true
```
zimbatm, line 35: This is not complete, it should only return true if `s.client.GetObject` doesn't return a 404.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// TODO: Implement GetFile for S3
```
zimbatm, line 38: ?

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *S3Store) GetFile(ctx context.Context, path string) (io.ReadCloser, error) {
+	// extend the struct to have the client
```
zimbatm, line 40: left-over comment

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+	object, err := s.client.GetObject(s.bucketName, path, minio.GetObjectOptions{})
+	if err != nil {
+			fmt.Println(err)
```
zimbatm, line 43: left-over debug

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+			return
+	}
```
zimbatm, line 45: the client should return the object here since it's a reader.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if u.Scheme != "s3" {
+		 fmt.Errorf("Invalid S3 URL")
+		 return
+	}
+
+	q := u.Query()
+
+	var useSSL bool
+    switch scheme := q.Get("scheme") {
```
zimbatm, line 64: indent

src/libstore/s3_binary_cache_store_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+		var re = regexp.MustCompile("([^&?]*?)=([^&?]*)")
+		var result = re.FindAllStringSubmatch(uri, -1)
+		var paramMap = make(map[string]string)
+
+		t.Errorf("%q", result)
+
+    if uri[0:5] != "s3://" {
+       t.Errorf("Sum was incorrect, got: %s, want: %s.", uri[0:4], "s3://")
+    }
```
zimbatm, line 22: it looks like an outdated test to me

Rizary:
Yeah my PR is merely for tracking purposes

Rizary:
@zimbatm I've made new commit today, please check.

src/libstore/s3_binary_cache_store.go
Rizary: do you mean if it is error, then we have to return `error` or `io.ReaderCloser`?

src/libstore/s3_binary_cache_store.go
Rizary: when we return in the `NewStoreReader`

src/libstore/s3_binary_cache_store.go
zimbatm: if the error is not nil then it `return nil, err`. Otherwise it would `return object, nil`.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,131 @@
...
+func (s *S3Store) getUri() string {
+	return "s3://" + s.bucketName
+}
+
+func (s *S3Store) FileExists(ctx context.Context, path string) (bool, error) {
+	found, err := s.client.BucketExists(ctx.Background(), s.bucketName)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
```
zimbatm, line 33: I wouldn't put that check here for two reasons:
* it doubles the number of requests that are issued when FileExists is called, which makes the function slower
* when GetObject is called on a bucket that doesn't exist, it will already return an error

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,131 @@
...
+	if found {
+		obj, err := s.client.GetObject(ctx, s.bucketName, path, minio.GetObjectOptions{})
+		if err != nil {
+			log.Fatalln(err)
```
zimbatm, line 37: In general, client libraries should avoid containing logging as it is hard to remove it. It's best to return the error object instead and let the parent deal with it.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,131 @@
...
+
+		stat, err := reader.Stat()
+		if err != nil {
+			log.Fatalln(err)
+		}
+
+		if _, err := io.CopyN(localFile, reader, stat.Size); err != nil {
+			log.Fatalln(err)
+			return
+		}
```
zimbatm, line 56: what is the purpose of getting the content locally?

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return false
```
zimbatm, line 59: did you build that code? this part shouldn't compile as it's missing the error part

Rizary: > did you build that code? this part shouldn't compile as it's missing the error part

I modified some test in the library.

src/libstore/s3_binary_cache_store.go
Rizary: To compare with the environment variable. If it's not set then it will return error

go.mod
```diff
@@ -4,5 +4,7 @@ go 1.12
 
 require (
 	cloud.google.com/go/storage v1.10.0
+	github.com/aws/aws-sdk-go v1.33.13
+	github.com/minio/minio-go/v7 v7.0.1
```
zimbatm, line 8: ultimately I would expect the code to only depend on one or the other libraries

src/libstore/s3_binary_cache_store.go
zimbatm: calling GetObject is enough to find out if a specific file exists or not, but you have to check the error response to see if it's a FileNotFound or something else.

src/libstore/doc.go
```diff
@@ -1,2 +0,0 @@
-// Package libstore implements a subset of the nix libstore API
-package libstore
```
zimbatm, line 2: this file is there by convention for godoc

src/libstore/s3_binary_cache_store.go
```diff
@@ -3,13 +3,12 @@ package libstore
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
+	awscred "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/minio/minio-go/v7"
+	miniocred "github.com/minio/minio-go/v7/pkg/credentials"
```
zimbatm, line 8: let's only use aws-sdk-go

src/libstore/s3_binary_cache_store.go
```diff
@@ -26,46 +25,30 @@ func (s *S3Store) getUri() string {
...
+	// found, err := s.client.BucketExists(ctx, s.bucketName)
+	// if err != nil {
+	// 	fmt.Println(err)
+	// 	return false, err
+	// }
+
+	// if found {
+	// 	fmt.Println(found)
+		_, err := s.GetFile(ctx, path)
 		if err != nil {
```
zimbatm, line 37: here you need to check what type of error is returned to filter:
* if it's an error because of file not found, return `false, nil`
* otherwise, return `false, err`

src/libstore/s3_binary_cache_store.go
```diff
@@ -26,46 +25,30 @@ func (s *S3Store) getUri() string {
...
-		if _, err := io.CopyN(localFile, reader, stat.Size); err != nil {
-			log.Fatalln(err)
-			return
-		}
-	}
-
-	return false
+		// defer obj.Close()
+	// }
+	return true, err
```
zimbatm, line 42: make it more obvious that err is always nil in that branch
```suggestion
	return true, nil
```

src/libstore/s3_binary_cache_store.go
```diff
@@ -26,46 +25,30 @@ func (s *S3Store) getUri() string {
...
 }
 
 func (s *S3Store) GetFile(ctx context.Context, path string) (io.ReadCloser, error) {
 	obj, err := s.client.GetObject(ctx, s.bucketName, path, minio.GetObjectOptions{})
 	if err != nil {
 		fmt.Println(err)
```
zimbatm, line 48: don't forget to remove that debugging line later

go.mod
zimbatm, line 8: please remove that dependency and then re-generate go.sum

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+	Scheme     string
+	BucketName string
+	Client     *s3.S3
+}
+
+func (s *S3Store) FileExists(path string) (bool, error) {
+
+	_, err := s.GetFile(path)
+	if err != nil {
+		return false, nil
```
zimbatm, line 27: here you need to switch on the error type and only return false if the error is of type `s3.ErrCodeNoSuchKey`

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+
+	obj, err := s.Client.GetObject(input)
+	if err != nil {
+    if aerr, ok := err.(awserr.Error); ok {
+        switch aerr.Code() {
+        case s3.ErrCodeNoSuchKey:
+            fmt.Println(s3.ErrCodeNoSuchKey, aerr.Error())
+        default:
+            fmt.Println(aerr.Error())
+        }
```
zimbatm, line 48: this code-switching is not needed here, just return the error

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+        // Print the error, cast err to awserr.Error to get the Code and
+        // Message from an error.
+        fmt.Println(err.Error())
```
zimbatm, line 52: All the Print statements should be removed from the library

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return nil, err
```
zimbatm, line 54: Please run `go fmt` on the code to have it indented properly.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+			S3ForcePathStyle: aws.Bool(true),
+
+		},
+	}))
+
+	svc := s3.New(sess)
+	return &S3Store{
+		Profile:    profile,
+		Region:     region,
+		Scheme:     scheme,
```
zimbatm, line 106: All these fields are not being used so you can remove them.

src/libstore/s3_binary_cache_store.go
```diff
@@ -0,0 +1,110 @@
...
+        // Message from an error.
+        fmt.Println(err.Error())
+    }
+    return nil, err
+	}
+
+	return obj.Body, nil // for now we return Object data with type blob
+}
+
+func NewStoreReader(uri string) (*S3Store, error) {
```
zimbatm, line 60:
```suggestion
func NewS3StoreReader(uri string) (*S3Store, error) {
```

tests/main.go
```diff
@@ -0,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+	"github.com/numtide/go-nix/src/libstore"
+)
+
+func main() {
+	r, err := libstore.NewStoreReader("s3://nar?region=eu-west-1&endpoint=http://127.0.0.1:9000&scheme=http")
```
zimbatm, line 9: use `libstore.NewBinaryCacheReader` and call NewS3StoreReader in that function

PR #6, opened 2021-02-21
nar/serialize.go: fix reader

flokli:
There's no guarantees on how many bytes the Read() method of an io.Reader will write.
We can't assume it to block until all bytes requested are there - it can
return before (especially if it's still receiving on the other side).

This adds a "hesitantReader", which behaves like this, a failing unit test, then switches serialize.go over to use io.ReadFull (which behaves like we want it to).

After all that is done, we switch over to encoding/binary instead of bitshifting manually to produce the (u)int64.

### DIFF
--- a/src/nar/serialize.go
+++ b/src/nar/serialize.go
@@ -63,15 +63,9 @@ const maxInt64 = 1<<63 - 1
 func readLongLong(r io.Reader) (int64, error) {
 	var num uint64
 	bs := make([]byte, 8, 8)
-	n, err := r.Read(bs)
-	if err != nil {
+	if _, err := io.ReadFull(r, bs); err != nil {
 		return 0, err
 	}
-	// FIXME: I think that io.Reader guarantees that
-	if n != 8 {
-		return 0, fmt.Errorf("expected to read 8 bytes, not %d", n)
-	}
-
 	num =
 		uint64(bs[0]) |
 			uint64(bs[1])<<8 |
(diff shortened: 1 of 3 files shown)


flokli:
Note the other `read*` methods suffer the same problem - I'll address that in a followup.

src/nar/serialize.go
```diff
@@ -1,10 +1,15 @@
 package nar
 
 import (
+	"encoding/binary"
 	"fmt"
 	"io"
 )
 
+const (
+	MaxInt64 = 1<<(64-1) - 1
```
zimbatm, line 10: `math.MaxInt64` also has that value. But maybe you didn't want to import the whole module just for the constant.

src/nar/serialize.go
```diff
@@ -61,32 +66,19 @@ func readPadding(r io.Reader, l int64) error {
 const maxInt64 = 1<<63 - 1
```
zimbatm, line 66: Is that a duplicate of the above? With a different value!

zimbatm (commented):
Good find!

src/nar/serialize.go
flokli: I think 64-1 also is 63, but yeah, I missed this constant being present :-)

Will update.

src/nar/serialize.go
flokli: I think the go compiler is pretty good in dead code elimination. I'll use `math.MaxInt64`, thanks!

PR #7, opened 2021-02-22
add wire package, remove serialize.go

flokli:
This introduces a `src/wire` package for the low-level field format, apparently used in both `.nar` files and nix-daemon communication. It not only supports more fields, but also the possibility to produce new fields:

```
func ReadBool(r io.Reader) (v bool, err error)
func ReadBytes(r io.Reader, max uint64) ([]byte, error)
func ReadString(r io.Reader, max uint64) (string, error)
func ReadUint64(r io.Reader) (n uint64, err error)
func WriteBool(w io.Writer, v bool) error
func WriteBytes(w io.Writer, buf []byte) error
func WriteString(w io.Writer, s string) error
func WriteUint64(w io.Writer, n uint64) error
```

When reading fields with arbitrary lengths, a maximum number of bytes needs to be specified.

It uses `ReadFull` consistently, so https://github.com/numtide/go-nix/pull/6#issuecomment-782905555 isn't a problem anymore for other field types.


src/wire/read.go
```diff
@@ -0,0 +1,78 @@
...
+// ReadBytes reads a bytes packet and returns a []byte of its contents
+// If the field exceeds the size passed via max, an error is returned
+// A bytes field starts with its size (8 bytes), then chunks of 8 bytes each.
+// Remaining bytes are padded with null bytes.
+func ReadBytes(r io.Reader, max uint64) ([]byte, error) {
+	// consume content length
+	contentLength, err := ReadUint64(r)
+	if err != nil {
+		return nil, err
+	}
```
zimbatm, line 58: it could return an empty byte slice if contentLength == 0

zimbatm (approved):
good stuff!

src/wire/read.go
flokli: done!

PR #8, opened 2021-02-25
Add CI

flokli:
```
commit dc0ad7b5753b89593c6b62556b86a21748c600b8
Author: Florian Klink <flokli@flokli.de>
Date:   Thu Feb 25 23:27:16 2021 +0100

    add GitHub Actions

commit aa21284c3d0c275296b480e17a97b5a99545bf09
Author: Florian Klink <flokli@flokli.de>
Date:   Thu Feb 25 23:02:17 2021 +0100

    flake.{lock,nix}, devshell.toml: remove
    
    This can be removed - the only thing that we need is a go compiler (to
    build and run unit tests).
    
    gopls is editor tooling, awscli and google-cloud-sdk is unused.
    minio/mc were used to provide a s3-compatible server for integration
    tests, but got removed in the previous commit, too
    
    Currently, this is simply a go libary, and the only requirement is a go
    compiler. We bring in different versions of it through CI, cause we want
    to test compatibility with (currently multiple) versions of it.

commit 0abba3540abfa8a4de8e2378f08203604ea42edc
Author: Florian Klink <flokli@flokli.de>
Date:   Thu Feb 25 23:33:55 2021 +0100

    remove integration test
    
     - This requires the user to have a working nix-daemon on their machine,
       to copy specific derivation from there.
     - This doesn't check for that derivation to be uploaded, but tries
       accessing the `nix-cache-info` file, that Nix might create when
       invoking nix copy. This isn't guaranteed to stay that way, and we
       shouldn't be integration testing nix copy.
     - It doesn't check for any file contents.
     - It doesn't return a nonzero exit code on error.
     - This requires the user to have a minio instance running on port 9000.
    
    Instead of doing all this, we should probably refactor the interface to
    allow passing in a mock client, and probe our implementation that way.
```


flokli:
GitHub Actions doesn't seem to be enabled in this repo.

I commented out line 5 and 6 in `.github/workflows/ci.yml`, and had a successfully run on my branch: https://github.com/flokli/go-nix/actions/runs/600964900

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+      matrix:
+        go: [ '1.14', '1.15', '1.16' ]
+    name: Build (Go ${{ matrix.go }})
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go }}
+    - name: go test -v ./...
+      run: cd src; go test -v ./...
```
zimbatm, line 22:
```suggestion
      run: go test -v ./...
```

PR #9, opened 2021-02-26
Fix nixbase32

flokli:
This is still WIP, as I'm a bit sure about the padding bits / `[En,De]codedLen`, and when the second error case in `DecodeString` can occur.


colemickens:
if it's helpful, I have an implementation here that I rely on https://github.com/colemickens/niche/tree/master/pkg/nixb32

flokli:
> if it's helpful, I have an implementation here that I rely on https://github.com/colemickens/niche/tree/master/pkg/nixb32

Thanks! However, this only treats encoding, not decoding, right?

flokli:
Using `strings.Builder` sounds like a good idea :-)

nixbase32/nixbase32.go
```diff
@@ -0,0 +1,81 @@
...
+		dst[i] |= byte(digit) << j
+
+		// calculate the "carry pattern"
+		carry := byte(digit) >> (8 - j)
+
+		// if we're at the end of dst…
+		if i == uint(len(dst)-1) {
+			// but have a nonzero padding, the encoding is invalid.
+			if carry != 0 {
+				return nil, fmt.Errorf("invalid encoding")
```
zimbatm, line 74: Maybe provide more information in the error message?

zimbatm (approved):
Looks good!

nixbase32/nixbase32.go
flokli: Hmmh, I'm not super sure.

Accodring to https://tools.ietf.org/html/rfc4648#section-3.2:

> Implementations MUST include appropriate pad characters at the end of encoded data unless the specification referring to this document explicitly states otherwise.

There's not really such a spec - this is mostly by looking at the "reference implementation".

I wouldn't want to expose internal terminology to consumers of that library. For "outside" consumers, "invalid encoding" seems to be enough of an error message, and people who really want to fiddle the nixbase32-specific details can easily go to this place, which now provides enough comment.

PR #13, opened 2021-06-02
libstore: add file protocol

zimbatm:
Imported from nar-serve

### DIFF
--- a/libstore/binary_cache.go
+++ b/libstore/binary_cache.go
@@ -35,6 +35,8 @@ func NewBinaryCacheReader(ctx context.Context, storeURL string) (BinaryCacheRead
 		return NewGCSBinaryCacheStore(ctx, u)
 	case "s3":
 		return NewS3BinaryCacheStore(u)
+	case "file":
+		return NewFileBinaryCacheStore(u), nil
 	default:
 		return nil, fmt.Errorf("scheme %s is not supported", u.Scheme)
 	}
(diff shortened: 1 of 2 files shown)


flokli:
@zimbatm what's the context for/purpose of these?

They seem to be pretty trivial, but I'm not sure if they should live in `go-nix` (at least not until they're documented).

zimbatm:
Is the S3 and GCS backend documented? That just adds support for pointing to a local copy of the cache, which can be useful for testing or other purposes. Nix also supports the file:// protocol, eg when using `nix copy`

flokli:
It's mostly to understand the purpose of these backends. I don't see where this is used right now - there's no consumers inside `go-nix`, and I already thought about removing the other two backends, as they literally just wrap different SDKs - and I'm not sure if the (undocumented) interface provides enough surface for its usages.

zimbatm:
my goal was to rebase nar-serve on top of go-nix. But I can also keep the libstore in nar-serve.

zimbatm:
Or move all of nar-serve into go-nix.

flokli:
It's mostly to understand the purpose of these backends.

I don't see where this is used right now - there's no consumers inside `go-nix`, and what it's supposed to do isn't really documented anywhere. From a distant look, it mostly wraps different SDKs and file operations on a storage backend and exposes it via an (undocumented) interface.

In fact, I already thought about removing the other two backends, as they aren't used anywhere.

Wonder if we should move this to `nar-serve` for the time being (which did vendor in go-nix anyways), or add some documentation on `libstore` and its intended usages?

PR #16, opened 2021-12-09
//nixbase32: expose alphabet

flokli:
This makes the alphabet publicly available, so consumers of go-nix can
use it.

### DIFF
--- a/nixbase32/nixbase32.go
+++ b/nixbase32/nixbase32.go
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-const alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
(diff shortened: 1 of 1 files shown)


nixbase32/nixbase32.go
```diff
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-const alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
+const Alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
```
zimbatm, line 8:
```suggestion
// Alphabet contains the list of valid characters for the Nix base32 alphabet.
const Alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
```

PR #17, opened 2021-12-09
//nar/narinfo/parser.go: handle super big .narinfo files

flokli:
Some .narinfo files (texlive-combined-full) have a lot of entries in
`References`, which will cause bufio.Scanner to error out with
bufio.ErrTooLong.

Increase the buffer size so things like this still parse.


nar/narinfo/parser.go
```diff
@@ -15,6 +15,13 @@ func Parse(r io.Reader) (*NarInfo, error) {
 	narInfo := &NarInfo{}
 	scanner := bufio.NewScanner(r)
 
+	// increase the buffer size.
```
zimbatm, line 18:
```suggestion
	// Increase the buffer size.
```

PR #21, opened 2022-02-20
feat: add flake/lint/fmt

jfroche:


jfroche:
This is still failing when using flake:

```
error: attribute 'mkShell' missing

       at /nix/store/ddaa56ll70qgpnk7y18lvdziarzsb6di-source/default.nix:17:14:

           16| {
           17|   devShell = devshell.mkShell {
             |              ^
           18|     packages = with nixpkgs; [
(use '--show-trace' to show detailed location information)
```
It must be obvious but I don't see why `devshell` doesn't contain `mkShell` when using flake...

jfroche:
We have to use `--impure` as we use `system ? builtins.currentSystem` in `default.nix` - is there an alternative ?

We cannot write in build dir in `checks`, this is fixed locally (by using $TMP) but cannot understand yet why it is failing in github action...

vendor/modules.txt
```diff
@@ -0,0 +1,20 @@
...
+## explicit
+# github.com/pmezard/go-difflib v1.0.0
+github.com/pmezard/go-difflib/difflib
+# github.com/stretchr/testify v1.6.1
+## explicit
+github.com/stretchr/testify/assert
+# gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
+## explicit
+# gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
+gopkg.in/yaml.v3
```
zimbatm: Is it necessary to vendor the dependencies? I'm not sure what the benefits are now that we have go modules.

zimbatm:
> We have to use --impure as we use system ? builtins.currentSystem in default.nix - is there an alternative ?

This shouldn't be needed as long as `system` is passed by the flake itself.

vendor/modules.txt
flokli: This is probably due to the `modules-download-mode: vendor` in `.golangci.yml`.

flokli:
Thanks for the cleanup! However, the vendoring of all go modules, and the introduction of nix flakes this is a bit too heavy-handed for my taste.

Please take a look at https://github.com/numtide/go-nix/pull/23, which moves `golangci-lint` to a Github Action, and cherry-picks your golangci-lint fixes.

jfroche:
moved to #23 
removing nix flake

PR #22, opened 2022-02-21
feat: add parser for derivation

jfroche:
depends on #21


jfroche:
Move `Value` into `Derivation`, drv file should contain 1 derivation

derivation/parser.go
```diff
@@ -0,0 +1,23 @@
+package derivation
+
+import (
+	"github.com/alecthomas/participle/v2"
+	"github.com/alecthomas/participle/v2/lexer"
+)
+
+var (
+	iniLexer = lexer.MustSimple([]lexer.Rule{
+		{Name: `NixPath`, Pattern: `"/nix/store/[a-zA-Z0-9+\-_?=][.a-zA-Z0-9+\-_?=]*"`},
```
flokli, line 10: Can we reuse bits from nixpath/nixpath.go here (possibly making them public)?

derivation/derivation.go
```diff
@@ -0,0 +1,63 @@
...
+	Path string   `'(' @NixPath ','`
+	Name []string `'[' ((@String ','?)* @String? )']' ')'`
+}
+
+type Env struct {
+	Key   string `'(' @String ','`
+	Value string `(@String|@NixPath)? ')'`
+}
+
+func ReadDerivation(filename string) (*Derivation, error) {
```
flokli, line 42: I'd keep all file io outside of this function. Just accept an io.Reader as function argument, and call `parser.Parse(` with it, instead of `parser.ParseString` below.

derivation/derivation.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}()
+
+	dat, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, err
+	}
+
+	drv := Derivation{}
+	err = parser.ParseString("", string(dat), &drv)
```
flokli, line 60: See above, use `io.Reader` and `parser.Parse`.

derivation/derivation.go
```diff
@@ -0,0 +1,63 @@
...
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		if err = file.Close(); err != nil {
+			log.Fatal(err)
+		}
+	}()
```
flokli, line 52: Right now, we mostly don't do much error handling when `defer`ing things that can return an error. I'd probably just `defer file.Close()`. However, if we change the whole function to accept a io.Reader (not io.ReadCloser) (and leave closing up to the caller), we don't need any `defer` at all.

derivation/derivation.go
```diff
@@ -0,0 +1,63 @@
...
+
+	dat, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, err
+	}
```
flokli, line 57: can be dropped

derivation/derivation.go
```diff
@@ -0,0 +1,63 @@
+// nolint: govet
+package derivation
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+type Derivation struct {
```
flokli, line 10: Please use the `parser:""` tag format, as described in https://github.com/alecthomas/participle#tag-syntax. This will allow co-existing with other tags - let's say if we want a JSON serialization and custom field names for it.

We can probably even add the `json:""` ones, to better explain how this maps to the fields showed in the `nix show-derivation` command.

derivation/parser.go
```diff
@@ -0,0 +1,23 @@
...
+		{Name: `DerivationPrefix`, Pattern: `Derive`},
+		{Name: `String`, Pattern: `"(?:\\.|[^"])*"`},
+
+		{Name: `Punct`, Pattern: `[,()\[\]]`},
+		{Name: "whitespace", Pattern: `[ \t\n\r]+`},
```
flokli, line 15: s/whitespace/Whitespace/?

derivation/parser.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/alecthomas/participle/v2/lexer"
+)
+
+var (
+	iniLexer = lexer.MustSimple([]lexer.Rule{
+		{Name: `NixPath`, Pattern: `"/nix/store/[a-zA-Z0-9+\-_?=][.a-zA-Z0-9+\-_?=]*"`},
+		{Name: `DerivationPrefix`, Pattern: `Derive`},
+		{Name: `String`, Pattern: `"(?:\\.|[^"])*"`},
+
+		{Name: `Punct`, Pattern: `[,()\[\]]`},
```
flokli, line 14: Maybe `Delim` instead of `Punct` is more clear?

flokli:
This should probably be rebased on top of #23.

flokli:
@jfroche do you have a chance to take a look at this?

jfroche:
> @jfroche do you have a chance to take a look at this?

Looking at it tonight

PR #23, opened 2022-02-22
.github/workflows: add golangci-lint.yml

flokli:
Alternative to #21.

It doesn't include the vendoring of all go modules, but instead drops the `modules-download-mode: vendor` line.

It also doesn't introduce 317 lines of Nix and flakes, which is a bit too heavy-handed for my taste.


nar/ls/list.go
```diff
@@ -8,12 +8,14 @@ import (
 	"github.com/numtide/go-nix/nar"
 )
 
+//nolint:revive
```
flokli, line 11: what is is complaining about here?

nar/ls/list.go
```diff
@@ -8,12 +8,14 @@ import (
...
 // LSRoot represents the .ls file root entry
 type LSRoot struct {
 	Version int `json:"version"`
 	Root    LSEntry
 }
 
+//nolint:revive
```
flokli, line 18: what is is complaining about here?

nar/narinfo/parser.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/numtide/go-nix/hash"
 )
 
+//nolint:gocyclo
```
flokli, line 13: what is is complaining about here?

nar/reader.go
```diff
@@ -20,9 +20,6 @@ const (
 	// maximum length for a single path element
 	// NAME_MAX is 255 on Linux
 	nameLenMax = 255
-	// maximum length for a relative path
-	// PATH_MAX is 4096 on Linux, but that includes a null byte
-	pathLenMax = 4096 - 1
```
flokli, line 25: Did we miss checking for this in the reader? We should probably re-introduce the check, instead of just dropping the (unused) constant.

nar/reader.go
```diff
@@ -250,9 +273,10 @@ func (nar *Reader) next() (*Header, error) {
 			if err != nil {
 				return nil, err
 			}
+
 			nar.level = append(nar.level, Entry)
 			// TODO: read the directory
-			//return h, nil
+			// return h, nil
```
flokli, line 279: What's missing here? This probably shouldn't be commented out…

nar/ls/list.go
flokli: Stuttering - fixed in 887c56a07f1d4692f6725f05d250b57bd231e280.

nar/narinfo/parser.go
flokli: Stuttering - fixed in 887c56a07f1d4692f6725f05d250b57bd231e280.

nar/ls/list.go
flokli: Stuttering - fixed in 887c56a07f1d4692f6725f05d250b57bd231e280.

nar/narinfo/parser.go
flokli: cyclomatic complexity 25 of func `Parse` is high (> 20). Probably fine to `nolint` that.

nar/reader.go
flokli: We did miss checking this, and checked the wrong thing for symlink targets. Pushed fixes in a834a0e01fc5bef5375085ef24b29a5798f4fab1 and b2adcab0fa8b4a7984d64d02fd38ad76ad7cc7b4.

nar/reader.go
flokli: I keep this as-is for now, until I digged into the parser more.

zimbatm (approved):

nar/header.go
```diff
@@ -10,7 +10,7 @@ import (
 type Header struct {
 	Type       EntryType // Typeflag is the type of header entry.
 	Name       string    // Name of the file entry
-	Linkname   string    // Target name of link (valid for TypeSymlink)
```
zimbatm, line 13: I think this was inherited from the Nix code base

flokli:
I trimmed down `.golangci.yml`, then went through https://golangci-lint.run/usage/linters/#disabled-by-default, and manually enabled most linters there that make sense, adressing the suggestions. PTAL.

zimbatm (approved):
LGTM

PR #25, opened 2022-03-14
feat(pkg/nar): rewrite

flokli:
This rewrites the NAR parser to a recursive version, running in its own
goroutine, which seems to be way easier to understand than the previous
iterative version.

Essentially, the structure we parse looks like this:

```
<Node>
 - ( type regular executable contents … )
 - ( type symlink target foo )
 - ( type directory
     entry ( name foo1 node <Node> )
     ...
   )
```

We use a channel to communicate with the outside world, so the iterative
interface can be preserved while internally having recursion instead of
manually managing stacks.

This also renames EntryType to NodeType. A Node is either a regular,
symlink or directory. `entry` only is a marker to list individual nodes
inside a directory.

~Before, there have been some changes in `pkg/wire`, improving test coverage and providing a Reader interface to bytes fields.~ Moved into #26


pkg/nar/header.go
```diff
@@ -8,11 +8,11 @@ import (
 // Header represents a single header in a NAR archive. Some fields may not
 // be populated depending on the Type.
 type Header struct {
-	Type       EntryType // Typeflag is the type of header entry.
-	Name       string    // Name of the file entry
-	LinkTarget string    // Target of symlink (valid for TypeSymlink)
-	Size       int64     // Logical file size in bytes
-	Executable bool      // Set to true for files that are executable
+	Path       string   // Path of the file entry, relative inside the NAR
```
zimbatm, line 11: If I remember correctly, this is the name of the entry. For example if you have a file under `bin/myexe`, you would have directory header with name "bin", and then a file with name "myexe". Path makes me think that the latter entry would contain `bin/myexe`.

zimbatm (approved):

PR #28, opened 2022-03-16
feat(pkg/nar/reader_test.go): also test closing the NAR reader

flokli:


pkg/nar/reader_test.go
```diff
@@ -234,7 +234,10 @@ func TestReader(t *testing.T) {
 	}
 
 	hdr, err := p.Next()
+	// expect to return io.EOF at the end, and no more headers
 	assert.Nil(t, hdr)
-	// expect to be finished
 	assert.Equal(t, io.EOF, err)
+
+	assert.NoError(t, p.Close(), nil, "closing the reader shouldn't error")
+	assert.NoError(t, p.Close(), nil, "closing the reader multiple times shouldn't error")
```
zimbatm, line 242: When an os.File is closed multiple times, it will err on the second call:

https://go.dev/play/p/LHDCVbhuF8J

flokli: I changed the test to ensure it doesn't panic. It seems there's not really a convention on how closing multiple times should behave.

NinjaTrappeur (approved):

NinjaTrappeur:
https://pkg.go.dev/os#File.Close

> Close will return an error if it has already been called. 

Seems in line with the `File` behavior.

flokli:
… and `archive/tar` returns nil on subsequent closures :shrug:. I guess the best thing to do is just check it doesn't panic ;-)

PR #30, opened 2022-03-17
feat(pkg/wire): update docs on Closing of a wire.BytesReader, don't seek to the end if ReadFull encounters a too big packet

flokli:


pkg/wire/read.go
```diff
@@ -65,17 +65,22 @@ func ReadBytes(r io.Reader) (uint64, io.ReadCloser, error) {
...
 	if err != nil {
 		return []byte{}, err
 	}
-	defer rd.Close()
 
 	if contentLength > max {
 		return nil, fmt.Errorf("content length of %v bytes exceeds maximum of %v bytes", contentLength, max)
 	}
 
+	defer rd.Close()
```
NinjaTrappeur, line 82: Nit: extract this in a second commit w/ some explanation about why you don't want to close the file if `contentLength > max` in the commit message?

pkg/wire/read_test.go
```diff
@@ -131,6 +131,13 @@ func TestReadString(t *testing.T) {
 	}
 
 	// exceeding max should error
-	_, err = wire.ReadString(bytes.NewReader(payloadFoo), 2)
+	rd := bytes.NewReader(payloadFoo)
```
NinjaTrappeur, line 134: nit: same: extract this in a second commit.

NinjaTrappeur (approved):
LGTM ~ how clean you want the git log to be.

PR #33, opened 2022-03-19
feat(pkg/nar): introduce NAR writer

flokli:
This introduces the writer component for `pkg/nar`, with a similar interface to `archive/tar`.

It also increases test coverage for the NAR reader, thanks to some of the fixtures created for the NAR writer tests.

There's probably some more places where we could have more test coverage, mostly premature closing of the NAR readers and writers, but this can be done as a follow-up.

### DIFF
--- a/pkg/nar/types.go
+++ b/pkg/nar/types.go
@@ -1,5 +1,12 @@
 package nar
 
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/nix-community/go-nix/pkg/nixpath"
+)
+
 const narVersionMagic1 = "nix-archive-1"
 
 // Enum of all the node types possible.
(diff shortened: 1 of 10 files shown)


pkg/nar/types.go
```diff
@@ -14,6 +21,8 @@ const (
 	TypeSymlink = NodeType("symlink")
 )
 
+var NodeNameRegexp = regexp.MustCompile(fmt.Sprintf("^%v$", nixpath.NameRe))
```
zimbatm, line 24: It seems like an implementation detail that should stay private to the module.

zimbatm (approved):
Aside from the goroutine business, which I am still unsure about, the rest looks good.

pkg/nar/types.go
flokli: Hmmh, it's also something that `pkg/nar/ls` could be using to verify the entry keys with. I'd prefer to keep it public.

PR #34, opened 2022-03-19
fix(pkg/nar): loosen node name restrictions

flokli:
Nix paths have much stronger requirements than regular paths. The only
thing we don't want to see is a / in the node name.

cc: @puckipedia

### DIFF
--- a/pkg/nar/header.go
+++ b/pkg/nar/header.go
@@ -1,7 +1,9 @@
 package nar
 
 import (
+	"fmt"
 	"os"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/nar/header.go
```diff
@@ -15,6 +17,44 @@ type Header struct {
...
+
+		if strings.ContainsAny(h.Path, "\u0000") {
+			return fmt.Errorf("path contains null bytes")
+		}
+	}
+
+	// Regular files and directories may not have LinkTarget set.
+	if h.Type == TypeRegular || h.Type == TypeDirectory {
+		if h.LinkTarget != "" {
+			if h.LinkTarget != "" {
```
andir, line 38: I think you can drop one of these checks as they are equivalent

pkg/nar/header.go
```diff
@@ -15,6 +17,44 @@ type Header struct {
...
+	if h.Type == TypeDirectory || h.Type == TypeSymlink {
+		if h.Size != 0 {
+			return fmt.Errorf("type is %v, but Size is not 0", h.Type.String())
+		}
+
+		if h.Executable {
+			return fmt.Errorf("type is %v, but Executable is true", h.Type.String())
+		}
+	}
+
```
andir, line 54: Currently empty link targets are accepts for `TypeSymlink`

pkg/nar/header.go
flokli: Oops, this was some reorganization error. I dropped it here and in ls/list.go.

pkg/nar/header.go
flokli: nice catch! fixed.

andir (approved):

PR #35, opened 2022-03-20
Introduce cmd/gonix, pkg/nar.DumpPath

flokli:
This introduces a CLI endpoint, exposing some of the functionality
provided inside pkg/.
    
Right now, it only exposes the functionality provided by the `nix nar
{cat,dump-path,ls}` commands in Nix, but its structure makes adding more
commands easy.

The logic used in the `nar dump-path` command was generic enough to be
moved to `pkg/nar`, and some tests were added.

`pkg/nar/header.go` received some love, and is now using less hardcoded
magic numbers. The `pkg/nar.Header`'s `FileInfo().Mode()` is getting used
to render the permission string in the `nar ls` subcommand.

### DIFF
--- a/cmd/gonix/nar/cat.go
+++ b/cmd/gonix/nar/cat.go
@@ -1,0 +1,55 @@
+package nar
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/nix-community/go-nix/pkg/nar"
+)
+
+type CatCmd struct {
+	Nar  string `kong:"arg,type='existingfile',help='Path to the NAR'"`
+	Path string `kong:"arg,type='string',help='Path inside the NAR, without leading slash'"`
+}
+
+func (cmd *CatCmd) Run() error {
+	f, err := os.Open(cmd.Nar)
+	if err != nil {
+		return err
+	}
+
+	nr, err := nar.NewReader(f)
+	if err != nil {
+		return err
+	}
+
+	for {
+		hdr, err := nr.Next()
+		if err != nil {
+			// io.EOF means we didn't find the requested path
+			if err == io.EOF {
+				return fmt.Errorf("requested path not found")
+			}
+			// relay other errors
+			return err
+		}
+
+		if hdr.Path == cmd.Path {
+			// we can't cat directories and symlinks
+			if hdr.Type != nar.TypeRegular {
+				return fmt.Errorf("unable to cat non-regular file")
+			}
+
+			w := bufio.NewWriter(os.Stdout)
+
+			_, err := io.Copy(w, nr)
+			if err != nil {
+				return err
+			}
+
+			return w.Flush()
+		}
+	}
+}
--- a/cmd/gonix/nar/ls.go
+++ b/cmd/gonix/nar/ls.go
@@ -1,0 +1,89 @@
+package nar
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/nix-community/go-nix/pkg/nar"
+)
+
+type LsCmd struct {
+	Nar       string `kong:"arg,type:'existingfile',help='Path to the NAR'"`
+	Path      string `kong:"arg,optional,type='string',default='',help='Path inside the NAR, without leading slash'"`
+	Recursive bool   `kong:"short='R',help='Whether to list recursively, or only the current level.'"`
+}
+
+// headerLineString returns a one-line string describing a header.
+// hdr.Validate() is assumed to be true.
+func headerLineString(hdr *nar.Header) string {
+	var sb strings.Builder
+
+	sb.WriteString(hdr.FileInfo().Mode().String())
+	sb.WriteString(" ")
+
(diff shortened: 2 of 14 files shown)


pkg/nar/dump.go
```diff
@@ -0,0 +1,129 @@
...
+
+		return err
+	}
+
+	err = nw.Close()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
zimbatm, line 34: `return nw.Close()` would be functionally equivalent

zimbatm (approved):
nice

NinjaTrappeur (changes requested):
There's something wrong with the `ls` command.

```sh
» ./gonix nar dump-path $(nix-build -E '(import <nixpkgs> {}).firefox') > firefox.nar
» ./gonix nar ls --recursive firefox.nar /  
» nix nar ls  firefox.nar /                                        
./bin
./lib
./share
```

We should extract that logic into `pkgs` and add some tests.

cmd/gonix/nar/ls.go
```diff
@@ -0,0 +1,89 @@
...
+			// io.EOF means we're done
+			if err == io.EOF {
+				return nil
+			}
+			// relay other errors
+			return err
+		}
+
+		// if the yielded path starts with the path specified
+		if strings.HasPrefix(hdr.Path, cmd.Path) {
```
NinjaTrappeur, line 73: `hdr.Path` will miss the leading `/` here, preventing `gonix ls $nar /(.*)` to match anything.

NinjaTrappeur:
Same issue for the cat command, we should either change the nar reader to prepend the top level `/`, either prepend it at the CLI boundary.

flokli:
@NinjaTrappeur the commands behave a bit different. There's no leading slash (and the listing one defaults to the empty string as prefix).

NinjaTrappeur:
Ah, ok. We should probably change the command name to clearly reflect the incompatibility with the Nix API then.

NinjaTrappeur (approved):

pkg/nar/dump.go
```diff
@@ -0,0 +1,129 @@
...
+	nw, err := NewWriter(w)
+	if err != nil {
+		return err
+	}
+
+	err = dumpPath(nw, path, "")
+	if err != nil {
+		// make sure the NAR writer is closed, so the underlying goroutine is stopped
+		// As we're already in error mode, discard a potentially returned error.
+		_ = nw.Close()
```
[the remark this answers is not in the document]
flokli: Good call! Did that, and the existing tests told me I didn't properly handle double closing in the writer :-D Fixed ;-)

flokli:
@NinjaTrappeur I added a help line to all arguments exposed `gonix nar {cat,ls} --help` now explicitly mention "without trailing slash" in the argument description.

cmd/gonix/nar/cat.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/nix-community/go-nix/pkg/nar"
+)
+
+type CatCmd struct {
+	Nar  string `kong:"arg,type='existingfile',help='Path to the NAR'"`
+	Path string `kong:"arg,type='string',help='Path inside the NAR, without trailing slash'"`
```
NinjaTrappeur, line 14:
```suggestion
	Path string `kong:"arg,type='string',help='Path inside the NAR, without heading slash'"`
```

cmd/gonix/nar/ls.go
```diff
@@ -0,0 +1,89 @@
...
+	"io"
+	"os"
+	"strings"
+
+	"github.com/nix-community/go-nix/pkg/nar"
+)
+
+type LsCmd struct {
+	Nar       string `kong:"arg,type:'existingfile',help='Path to the NAR'"`
+	Path      string `kong:"arg,optional,type='string',default='',help='Path inside the NAR, without trailing slash'"`
```
NinjaTrappeur, line 14:
```suggestion
	Path      string `kong:"arg,optional,type='string',default='',help='Path inside the NAR, without leading slash'"`
```

pkg/nar/dump.go
```diff
@@ -0,0 +1,123 @@
...
+			// If it's executable by the user, it'll become executable.
+			// This matches nix's dump() function behaviour.
+			Executable: fi.Mode()&syscall.S_IXUSR != 0,
+		})
+		if err != nil {
+			return err
+		}
+
+		// open the file
+		f, err := os.Open(p)
```
NinjaTrappeur, line 103: Maybe we should consider a `defer f.Close()` to prevent FD exhaustion here: we might end up opening a bunch of files in a single GC interval when dumping a large tree.

cmd/gonix/nar/cat.go
flokli: I assume you meant `leading`.

pkg/nar/dump.go
flokli: :+1:

PR #37, opened 2022-04-30
pkg/derivation: Don't rebuild parser for every call to ReadDerivation

adisbladis:
Measured performance with the following program:
``` go
package main

import (
	"bytes"
	"fmt"
	"io"
	"os"
	"sync"
	"time"
)

func main() {
	var wg sync.WaitGroup

	derivationFile, err := os.Open("./test/testdata/m5j1yp47lw1psd9n6bzina1167abbprr-bash44-023.drv")
	if err != nil {
		panic(err)
	}

	b, err := io.ReadAll(derivationFile)
	if err != nil {
		panic(err)
	}

	start := time.Now()

	for i := 1; i < 10000; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			reader := bytes.NewReader(b)

			_, err := derivation.ReadDerivation(reader)
			if err != nil {
				panic(err)
			}
		}()
	}
	wg.Wait()

	fmt.Println(time.Since(start))
}
```

Before:
``` shell
$ /run/current-system/sw/bin/time -f "%M" go run tst.go
1.026543578s
157200  # Max RSS
```

After:
``` shell
$ /run/current-system/sw/bin/time -f "%M" go run tst.go
432.034462ms
80816  # Max RSS
```

The max RSS numbers are a bit unreliable, but enough to confidently say that memory usage has reduced by quite a bit.

-----
cc @jfroche

### DIFF
--- a/pkg/derivation/derivation.go
+++ b/pkg/derivation/derivation.go
@@ -37,23 +37,21 @@ type Env struct {
 	Value string `parser:"(@String|@NixPath)? ')'"`
 }
 
-func ReadDerivation(reader io.Reader) (*Derivation, error) {
-	drv := Derivation{}
-	iniLexer := lexer.MustSimple([]lexer.Rule{
+// nolint:gochecknoglobals
+var parser = participle.MustBuild(&Derivation{},
+	participle.Lexer(lexer.MustSimple([]lexer.Rule{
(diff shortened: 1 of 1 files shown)


jfroche (approved):
Nice improvement. Thank you !

PR #38, opened 2022-04-30
pkg/derivation: Use WriteDerivation(io.Writer) for serialising derivations

adisbladis:
To turn a parsed derivation back into a derivation file.


pkg/derivation/encode.go
```diff
@@ -0,0 +1,59 @@
+package derivation
+
+import (
+	"strings"
+)
+
+func quoteString(s string) string {
```
flokli, line 7: right now, this only encloses the string in double quotes, and doesn't do any escaping. Can you add a comment to the function header, explaining why no escaping is needed?

flokli (commented):
The linter complains, and I'd love if `quoteString`/`encodeArray` would have some comments.

flokli:
> The linter complains, and […]

Linter was fixed in https://github.com/nix-community/go-nix/pull/39

jfroche:
Great work, thank you !

We might want to do the same as we do in nix where the string representation of a derivation is the actual output path ?
And have instead a `WriteDerivation` with `io.Writer` ?

adisbladis:
> We might want to do the same as we do in nix where the string representation of a derivation is the actual output path ?

That sounds like it might be a good idea. At the same time I think `Derivation` should have a `GetOutput(output string)`.

>  And have instead a WriteDerivation with io.Writer ?

I quite like the symmetry of that together with `ReadDerivation`!

jfroche (approved):
LGTM :+1:

flokli (approved):

flokli:
> Great work, thank you !
> 
> We might want to do the same as we do in nix where the string representation of a derivation is the actual output path ? And have instead a `WriteDerivation` with `io.Writer` ?

Thinking about this again, it might be nicer if `Derivation.String()` would return the (calculated) path to the `.drv` file itself.

This requires a bunch of hashing to be implemented in `go-nix` (see https://github.com/nix-community/go-nix/issues/41) for details, but would be what the nix repl returns when calling `builtins.derivation`:

```
nix-repl> builtins.derivation { name = "foo"; builder = ":"; system = ":"; }
«derivation /nix/store/xa96w6d7fxrlkk60z1fmx2ffp2wzmbqx-foo.drv»
```

We can already explicitly request the outputs we want, but `Derivation.String()` not returning the derivation path/name feels a bit weird.

flokli:
I spoke with @adisbladis a bit. Just by looking at the contents of the `Derivation` struct we have here, it's not possible to calculate the hash as soon as inputDrvs are present. You need to know more about the other derivations that are referred.

jfroche:
@flokli I don't see why it is impossible to calculate if we have access to the referred input derivations.

adisbladis:
@jfroche That was the point. You'll need access to the entire derivation graph. You cannot just do it as a pure computation given a single Derivation struct.

jfroche:
Right ok, I was not sure what you meant by "impossible"

PR #43, opened 2022-05-28
pkg/nar/writer: Compare order of NAR members correctly

adisbladis:
The current implementation breaks when the string representation for
the _full path_ is not lexicographically ordered.

This breaks when two nested directories share a prefix such as:
- /cmd/structlayout/main.go
- /cmd/structlayout-optimize

The full path for these strings is not in correct lexicographical
order, but each component is.

### DIFF
--- a/pkg/nar/util.go
+++ b/pkg/nar/util.go
@@ -7,3 +7,28 @@ import "strings"
 func IsValidNodeName(nodeName string) bool {
 	return !strings.Contains(nodeName, "/") && !strings.ContainsAny(nodeName, "\u0000")
 }
+
+// PathIsLexicographicallyOrdered checks if two paths are lexicographically ordered component by component.
+func PathIsLexicographicallyOrdered(path1 string, path2 string) bool {
+	if path1 > path2 {
+		path1Segments := strings.Split(path1, "/")
+		path2Segments := strings.Split(path2, "/")
+
+		// n is the lower number of segments of the two paths.
+		var n int
+		if len(path1Segments) < len(path2Segments) {
+			n = len(path1Segments)
+		} else {
+			n = len(path2Segments)
+		}
+
+		// check all segments individually
+		for i := 0; i < n; i++ {
+			if !(path1Segments[i] <= path2Segments[i]) {
+				return false
+			}
+		}
+	}
+
+	return true
+}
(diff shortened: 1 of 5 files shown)


flokli (commented):
Thanks! Can you add a regression test for this?

pkg/nar/util.go
```diff
@@ -7,3 +7,26 @@ import "strings"
 func IsValidNodeName(nodeName string) bool {
 	return !strings.Contains(nodeName, "/") && !strings.ContainsAny(nodeName, "\u0000")
 }
+
+// isLexicographicallyOrdered checks if two paths are lexicographically ordered component by component.
+func isLexicographicallyOrdered(path1 string, path2 string) bool {
```
flokli, line 12: Maybe you can also call this PathisLexicographicallyOrdered, and just write a unit test for this, covering the various branches.

pkg/nar/util.go
```diff
@@ -7,3 +7,26 @@ import "strings"
...
+		} else {
+			n = len(path2Split)
+		}
+
+		for i := 0; i < n; i++ {
+			if !(path1Split[i] <= path2Split[i]) {
+				return false
+			}
+		}
+	}
```
flokli, line 29:
```suggestion
		path1Segments := strings.Split(path1, "/")
		path2Segments := strings.Split(path2, "/")

    // n is the lower number of segments of the two paths.
		var n int
		if len(path1Segments) < len(path2Segments) {
			n = len(path1Segments)
		} else {
			n = len(path2Segments)
		}

    // check all segments individually
		for i := 0; i < n; i++ {
			if !(path1Segments[i] <= path2Segments[i]) {
				return false
			}
		}
	}
```

adisbladis:
> Thanks! Can you add a regression test for this?

Done!

pkg/nar/util.go
adisbladis: Done!

flokli:
Thanks!

pkg/nar/util.go
```diff
@@ -7,3 +7,28 @@ import "strings"
...
+		} else {
+			n = len(path2Segments)
+		}
+
+		// check all segments individually
+		for i := 0; i < n; i++ {
+			if !(path1Segments[i] <= path2Segments[i]) {
+				return false
+			}
+		}
```
zimbatm, line 30: If somebody wants to play with performance, a straight for loop over each character is probably more efficient. It would avoid the allocations and second pass over the string that strings.Split causes.

adisbladis: Good catch! Implemented in https://github.com/nix-community/go-nix/pull/44.

PR #44, opened 2022-05-30
pkg/nar: Avoid allocations in PathIsLexicographicallyOrdered

adisbladis:


pkg/nar/util.go
```diff
@@ -10,25 +10,30 @@ func IsValidNodeName(nodeName string) bool {
...
-		path1Segments := strings.Split(path1, "/")
-		path2Segments := strings.Split(path2, "/")
-
-		// n is the lower number of segments of the two paths.
-		var n int
-		if len(path1Segments) < len(path2Segments) {
-			n = len(path1Segments)
-		} else {
-			n = len(path2Segments)
+	if path1 < path2 {
```
zimbatm, line 13:
```suggestion
	if path1 <= path2 {
```
Another micro-optim

PR #46, opened 2022-05-30
pkg/nar: Add DumpPathFilter

adisbladis:
This function adds source filtering capabilities to DumpPath.
The interface of this mimics the behaviour of the Nix builtin filterSource.

### DIFF
--- a/pkg/nar/dump.go
+++ b/pkg/nar/dump.go
@@ -8,9 +8,21 @@ import (
 	"syscall"
 )
 
+// SourceFilterFunc is the interface for creating source filters.
+// If the function returns true, the file is copied to the Nix store, otherwise it is omitted,
+// this mimics the behaviour of the Nix function builtins.filterSource.
+type SourceFilterFunc func(path string, nodeType NodeType) bool
+
 // DumpPath will serialize a path on the local file system to NAR format,
 // and write it to the passed writer.
 func DumpPath(w io.Writer, path string) error {
+	return DumpPathFilter(w, path, nil)
+}
+
+// DumpPathFilter will serialize a path on the local file system to NAR format,
+// and write it to the passed writer, filtering out any files where the filter
+// function returns false.
+func DumpPathFilter(w io.Writer, path string, filter SourceFilterFunc) error {
 	// initialize the nar writer
 	nw, err := NewWriter(w)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/nar/dump.go
```diff
@@ -8,9 +8,20 @@ import (
 	"syscall"
 )
 
+// SourceFilterFunc is the interface for creating source filters,
+// it mimics the behaviour of the Nix builtin filterSource.
```
flokli, line 12: Can you describe behaviour in a bit more detail?

 - Keep on true/omit on false
 - The input values into the filter function. I assume `path` would be a more fitting name for the first argument?

pkg/nar/dump.go
```diff
@@ -118,6 +143,8 @@ func dumpPath(nw *Writer, path string, subpath string) error {
 		}
 
 		return nil
+
+	case TypeUnknown:
```
zimbatm, line 147: Move the `return fmt.Errorf("invalid mode for file %v", p)` from below?

pkg/nar/dump.go
```diff
@@ -8,9 +8,20 @@ import (
 	"syscall"
 )
 
+// SourceFilterFunc is the interface for creating source filters,
+// it mimics the behaviour of the Nix builtin filterSource.
+type SourceFilterFunc func(name string, nodeType NodeType) bool
```
flokli, line 13:
```suggestion
type SourceFilterFunc func(path string, nodeType NodeType) bool
```

pkg/nar/header.go
```diff
@@ -92,6 +92,8 @@ func (fi headerFileInfo) Mode() fs.FileMode {
 		mode |= (syscall.S_IXUSR | syscall.S_IXGRP | syscall.S_IXOTH)
 	case TypeSymlink:
 		mode = fs.ModePerm | fs.ModeSymlink
+	case TypeUnknown:
```
flokli, line 95: Add a comment why this is impossible to construct through NAR reader, only by manually populating structs.

Also, please add a check for it to the `Header.Validate()` function, so it's possible to check without panic'ing.

flokli:
Ah, and while at it, please also extend https://github.com/nix-community/go-nix#pkgnar, to also mention DumpPathFilter.

pkg/nar/dump.go
adisbladis: I tried that first, but then we'll still need to return the error at the end of the function to appease the compiler.
To avoid duplication I just let it fall through.

adisbladis: `golangci-lint` requires all enum members to be present, so we're back to having to duplicate the error return in that case.

adisbladis: golangci-lint requires the switch to be exhaustive (ie contain `TypeUnknown`), so we're back to duplicating the error return or opt out of the linter.
I would rather just stick the error at the end of the function.

adisbladis:
All comments so far addressed.

PR #47, opened 2022-06-01
pkg/derivation: Add Validate() method

### LINKED ISSUE #41: pkg/derivation: add `Validate()` method
There's a bunch of invalid derivations that you can create, either by populating the structs wrongly, or by feeding the parser with an invalid serialization.

Like in other places in `go-nix`, we should have a `Validate()` function, that provides some basic sanity checks, such as:

 - [x] At least one output defined
 - [x] Nix paths matching the regexes defined in `pkg/nixpath`
 - [ ] The output hashes actually being calculated correctly (https://comono.id/posts/2020-03-20-how-nix-instantiation-works/, https://nixos.org/guides/nix-pills/nix-store-paths.html)

Like in `pkg/nar`, we should call that `Validate()` function during the deserializaiton phase (`pkg/derivation.ReadDerivation`).

cc @jfroche @adisbladis

adisbladis:
Implements parts of #41

### DIFF
--- a/pkg/derivation/derivation.go
+++ b/pkg/derivation/derivation.go
@@ -20,6 +20,66 @@ type Derivation struct {
 	EnvVars          []Env             `json:"env" parser:"'[' ((@@ ','?)* (@@)? )']'')'"`
 }
 
+func (d *Derivation) Validate() error {
+	if len(d.Outputs) == 0 {
+		return fmt.Errorf("at least one output must be defined")
+	}
+
+	for i, o := range d.Outputs {
+		err := o.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating output '%s': %w", o.Content, err)
+		}
+
+		if i > 0 && o.Content < d.Outputs[i-1].Content {
+			return fmt.Errorf("invalid output order: %s < %s", o.Content, d.Outputs[i-1].Content)
+		}
+	}
+
+	for i, id := range d.InputDerivations {
+		err := id.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating input derivation '%s': %w", id.Path, err)
+		}
+
+		if i > 0 && id.Path < d.InputDerivations[i-1].Path {
+			return fmt.Errorf("invalid input derivation order: %s < %s", id.Path, d.InputDerivations[i-1].Path)
+		}
+	}
+
+	for i, is := range d.InputSources {
+		_, err := nixpath.FromString(is)
+		if err != nil {
+			return fmt.Errorf("error validating input source '%s': %w", is, err)
+		}
+
+		if i > 0 && is < d.InputSources[i-1] {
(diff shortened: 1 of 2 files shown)


pkg/derivation/derivation.go
```diff
@@ -20,6 +20,50 @@ type Derivation struct {
 	EnvVars          []Env             `json:"env" parser:"'[' ((@@ ','?)* (@@)? )']'')'"`
 }
 
+func (d *Derivation) Validate() error {
+	if len(d.Outputs) == 0 {
+		return fmt.Errorf("at least one output must be defined")
+	}
+
+	for _, o := range d.Outputs {
```
flokli, line 28: Do we need to validate the order of elements being sorted?

pkg/derivation/derivation.go
```diff
@@ -20,6 +20,50 @@ type Derivation struct {
...
+		err := o.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating output: %w", err)
+		}
+	}
+
+	for _, id := range d.InputDerivations {
```
flokli, line 35: same, can we check for a certain order of inputs?

pkg/derivation/derivation.go
```diff
@@ -20,6 +20,50 @@ type Derivation struct {
...
+		err := id.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating input derivation: %w", err)
```
flokli, line 38: specify which input we're erroring about

pkg/derivation/derivation.go
```diff
@@ -20,6 +20,50 @@ type Derivation struct {
...
+	}
+
+	if d.Builder == "" {
+		return fmt.Errorf("required attribute 'builder' missing")
+	}
+
+	for _, e := range d.EnvVars {
+		err := e.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating env var: %w", err)
```
flokli, line 60: env var key should probably be mentioned in the error message

pkg/derivation/derivation.go
```diff
@@ -20,6 +20,50 @@ type Derivation struct {
...
+func (d *Derivation) Validate() error {
+	if len(d.Outputs) == 0 {
+		return fmt.Errorf("at least one output must be defined")
+	}
+
+	for _, o := range d.Outputs {
+		err := o.Validate()
+		if err != nil {
+			return fmt.Errorf("error validating output: %w", err)
+		}
```
flokli, line 32: There's a deterministic way to come up with the hashes of the output path, by looking at the drv.

Check page 94ff of https://edolstra.github.io/pubs/phd-thesis.pdf.

As we don't have the code for doing the hashing yet, we should log a FUTUREWORK here.

flokli: added FUTUREWORK in 332b5a18f086b9075ea0b5c0da4355c30e1a66c1.

PR #50, opened 2022-06-01
Fix derivation parser (builder field)

adisbladis:
I don't know if this fix is strictly correct since I'm not too familiar with the parser library.

cc @jfroche


jfroche (approved):
Nice !

adisbladis:
Fwiw: This (and other `bootstrap-tools` derivations) were the only ones on my local system that failed parsing.

PR #54, opened 2022-06-03
Implement derivation output hash calculation

adisbladis:


pkg/derivation/derivation.go
```diff
@@ -81,29 +86,163 @@ func (d *Derivation) Validate() error {
...
+
+func (d *Derivation) Name() string {
+	for _, e := range d.EnvVars {
+		if e.Key == "name" {
+			return e.Value
+		}
+	}
+
+	// TODO: Maybe panic or change type sig to (string, error)?
+	return ""
```
flokli, line 105: `panic` here, and return an error in all parser codepaths - which in our case, would be best in the `Validate()` function (so people populating structs manually could still check if through `Validate()` without being afraid of panics)

flokli: Ah, and add a comment next to the panic statement that this is verified in the `Validate` method, so can't really happen on validated inputs.

pkg/derivation/derivation.go
```diff
@@ -81,29 +86,163 @@ func (d *Derivation) Validate() error {
...
+}
+
+func (d *Derivation) OutputPaths(store KVStore) (map[string]string, error) {
```
flokli, line 108: function comment

pkg/derivation/kv.go
```diff
@@ -0,0 +1,95 @@
...
+	return &InputSubstKV{
+		memo:  make(map[string][]byte),
+		store: store,
+	}
+}
+
+func (kv *InputSubstKV) Get(key string) ([]byte, error) {
+	value, ok := kv.memo[key]
+
+	if !ok {
```
flokli, line 45: I guess we can just do an early return if okay, and then indent the whole rest one level less.

pkg/derivation/derivation.go
```diff
@@ -81,29 +86,163 @@ func (d *Derivation) Validate() error {
...
+		}
+		digest := h.Sum(nil)
+
+		outputs[o.Name] = filepath.Join(nixpath.StoreDir, nixbase32.EncodeToString(compressHash(digest, 20))+"-"+outputSuffix)
+	}
+
+	return outputs, nil
+}
+
+// FixedOutput - Returns the fixed output any is found, otherwise returns nil.
```
flokli, line 168: Should this be public api? What are the consumers?

What happens if there's more than one output where `HashAlgorithm` is the empty string?

pkg/derivation/derivation.go
```diff
@@ -81,29 +86,182 @@ func (d *Derivation) Validate() error {
...
-	for i, o := range d.Outputs {
-		outputs[i] = encodeArray('(', ')', true, []byte(o.Name), []byte(o.Path), []byte(o.HashAlgorithm), []byte(o.Hash))
+	{
+		for i, o := range d.Outputs {
+			var path []byte
+			if maskOutputs {
+				path = []byte{}
+			} else {
+				path = []byte(o.Path)
+			}
```
flokli, line 208: It's probably less code and allocs if you just repeat L210 two times, for both cases.

pkg/derivation/derivation.go
```diff
@@ -81,29 +86,182 @@ func (d *Derivation) Validate() error {
...
+
+				path = append([]byte{'"'}, path...) // TODO: Inefficient
+				path = append(path, '"')
+			} else {
+				path = quoteString(in.Path)
+			}
+
 			names := encodeArray('[', ']', true, stringsToBytes(in.Name)...)
-			inputDerivations[i] = encodeArray('(', ')', false, quoteString(in.Path), names)
+			inputDerivations[i] = encodeArray('(', ')', false, path, names)
```
flokli, line 233: same as above, probably easier to just write two different blocks setting `inputDerivations[i]`, depending on whether `actualInputs` is nill or not.

pkg/derivation/kv.go
```diff
@@ -0,0 +1,94 @@
...
+
+type FSKVStore struct{}
+
+func (kv *FSKVStore) Get(key string) ([]byte, error) {
+	return os.ReadFile(key)
+}
+
+func (kv *FSKVStore) Set(key string, value []byte) error {
+	return fmt.Errorf("writing not supported")
+}
```
flokli, line 26: I'd probably start with a simple hashmap.

Reading from random places in the filesystem isn't good style, especially when used in testcases.

flokli: Also, this will give us the opportunity to think about how things are added into the KV store. Is it ReadDerivation that'd get a reference passed in?

adisbladis: I needed the file system implementation to try this out on my local Nix store.

pkg/derivation/derivation.go
flokli: I think this is related to 95f413a6e5edc8b88e56e144dc2e83676d8551a7.

pkg/derivation/kv.go
adisbladis: We might want to limit it to read from `nixpath.StoreDir` or some passed in prefix?

pkg/derivation/kv.go
```diff
@@ -0,0 +1,94 @@
...
+
+func NewInputSubstKV(store KVStore) *InputSubstKV {
+	return &InputSubstKV{
+		memo:  make(map[string][]byte),
+		store: store,
+	}
+}
+
+func (kv *InputSubstKV) Get(key string) ([]byte, error) {
```
flokli, line 42: This doesn't seem to return what you passed in, but does a whole bunch of calculation on what it from the underlying KV.

It should probably be named differently - and I'd argue we want to do that calculation once, while calling Set(), and then just return the cached result.

pkg/derivation/kv.go
flokli: Yes, for `FSKVStore`. But I think we need a different structure:

I can see some sort of component (let's call it `DrvStore` for now) that stores 2 things in a K/V structure:

 - a map from .drv path to .drv file contents
 - a map from .drv path to (intermediate) hashes.

The FSKVStore would be one possible candidate for the first concern, if you want to read .drv files on the local machine (and restricting that to `nixpath.StoreDir`, and only legal `.drv` paths definitely makes sense).

However, another valid usecase is parsing `.drv` files not on disk, but parsing .drv files from a binary cache for example.
For this, we want to have an alternative implementation, that might just be living in memory.

The second KV holds intermediate hashes. It is what's queried frequently "from the outside", and internally for hash calculation. We probably need an interface to insert a-terms into a `DrvStore` too (in the case of memory K/V, we certainly do), so there we can precompute intermediate hashes on insert.

In the case of a `DrvStore` backed the local filesystem (`FSKVStore`), we probably don't want to fail on inserting (because it's Nix that owns these files), but we should still populate a cache of intermediate hashes, whenever there's a query about a certain .drv.

flokli:
Closing in favor of https://github.com/nix-community/go-nix/pull/58.

PR #55, opened 2022-06-04
pkg/derivation: Add an allocation free derivation parser 

adisbladis:
This quite drastically improves performance and fixes a bug I noticed in regards to how the current parser deals with escaped strings.
Currently the derivation parser repeats backslashes, and the derivation writer works around that by rewriting the double escaping.

My main motivation though was performance which can be demonstrated by the following program:
``` go
package main

import (
	"bytes"
	"fmt"
	"io"
	"os"
	"testing"
	"time"

)

func main() {
	f, err := os.Open("/nix/store/5zhfmj5j8b326g6pxlc4ky6i2il20avp-jq-1.6.drv")
	if err != nil {
		panic(err)
	}

	b, err := io.ReadAll(f)
	if err != nil {
		panic(err)
	}

	start := time.Now()

	for i := 0; i < 10_000; i++ {
		_, err := derivation.ReadDerivation(bytes.NewReader(b))
		if err != nil {
			panic(err)
		}
	}

	fmt.Println(time.Since(start))

	fmt.Println("Allocs:", int(testing.AllocsPerRun(10, func() {
		_, err := derivation.ReadDerivation(bytes.NewReader(b))
		if err != nil {
			panic(err)
		}
	})))

}
```

- Before
```
4.581642959s
Allocs: 7829
```

- After
```
299.002489ms
Allocs: 169
```

A runtime improvement of about 15x and about 45x fewer allocations!

cc @jfroche


adisbladis:
I also tried this on `` (I had to adjust the number of iterations down to 5 to get it to complete in a reasonable time):

- Before
```
37.573942663s
Allocs: 1575726
```

- After
```
48.310899ms
Allocs: 27138
```

So roughly 750x faster in terms of time and a 60x improvement in terms of allocations.

pkg/derivation/encode.go
```diff
@@ -2,23 +2,11 @@ package derivation
...
-	"\"", "\\\"",
 )
 
 // Escapes user provided values such as derivation attributes.
 // These may contain special characters such as newlines, tabs, backslashes and so on.
 func escapeString(s string) []byte {
-	s = stringEscaper.Replace(s)
-
 	return quoteString(s)
 }
```
flokli: You can remove `escapeString` here, and just `quoteString` exclusively in `pkg/derivation/derivation.go`.

flokli (approved):

PR #56, opened 2022-06-06
pkg/derivation: align Derivation struct with json format

flokli:
A lot of the struct structure has been designed to work with the parser
generator we used.

Since https://github.com/nix-community/go-nix/pull/55 we have our own parser, so we can be a bit more flexible with
how the structure of the Derivation struct looks like.

Use the structure that's exposed when invoking `nix show-derivation`,
and peeking at the JSON there.

Outputs, InputDerivations and Env are maps, not lists with odd order
requirements. The Parser takes care of ensuring the order was correct
when parsing it in, and the encoder outputs them in the way things
should be sorted.

This makes the Validate() method much smaller, as there's simply less
invalid states someone manually populating the struct could produce.

### DIFF
--- a/pkg/derivation/derivation.go
+++ b/pkg/derivation/derivation.go
@@ -108,12 +108,6 @@ func (d *Derivation) Validate() error {
 	return nil
 }
 
-// String returns the path of the derivation.
-func (d *Derivation) String() string {
-	// TODO: replace with drv path
-	return "TODO"
-}
-
 type Output struct {
 	Path          string `json:"path"`
 	HashAlgorithm string `json:"hashAlgo"`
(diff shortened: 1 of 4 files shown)


adisbladis (approved):

pkg/derivation/derivation.go
```diff
@@ -64,121 +88,43 @@ func (d *Derivation) Validate() error {
...
-	return err
+	return nil
 }
 
-// String returns the default (first) output path.
+// String returns the path of the derivation.
 func (d *Derivation) String() string {
-	return d.Outputs[0].Path
+	// TODO: replace with drv path
+	return "TODO"
```
adisbladis, line 114: Maybe remove this for now? It's behaviour is currently incorrect anyway, and returning `TODO` is not useful.

PR #57, opened 2022-06-06
pkg/derivation: add DrvPath() function

flokli:
This returns the path of a Drv, just by looking at its structure.

### DIFF
--- a/pkg/derivation/derivation_test.go
+++ b/pkg/derivation/derivation_test.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/nix-community/go-nix/pkg/derivation"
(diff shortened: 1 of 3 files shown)


pkg/derivation/util.go
```diff
@@ -0,0 +1,15 @@
+package derivation
+
+// CompressHash takes an arbitrary long sequence of bytes (usually a hash digest),
+// and returns a sequence of bytes of length newSize.
+// It's calculated by rotating through the bytes in the output buffer (zero-initialized),
+// and XOR'ing with each byte in the passed input
+// It consumes 1 byte at a time, and XOR's it with the current value in the output buffer.
+func CompressHash(input []byte, outputSize int) []byte {
```
adisbladis, line 8: This should probably be moved to the `hash` package.

adisbladis (approved):

PR #58, opened 2022-06-06
pkg/derivation: implement Output Path calculation

flokli:
Refactored version of https://github.com/nix-community/go-nix/pull/54.

Closes https://github.com/nix-community/go-nix/issues/41.

### DIFF
--- a/pkg/derivation/output_path.go
+++ b/pkg/derivation/output_path.go
@@ -1,0 +1,90 @@
+package derivation
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+)
+
+// OutputPaths returns a map of output names to the calculated store paths.
+// For more than trivial derivations, calculating these requires some
+// substitutions of other referred InputDerivations, so a store needs to be passed,
+// that can be used to look up other Derivations.
+func (d *Derivation) OutputPaths(store Store) (map[string]string, error) {
+	derivationName, ok := d.Env["name"]
+	if !ok {
+		// asserted by Validate
+		panic("env 'name' not found")
+	}
+
+	// we only want to call partial drv hash if we're not a fixed output
+	var partialDrvHash string
+
+	if fixed := d.GetFixedOutput(); fixed == nil {
+		var buf bytes.Buffer
+
+		err := d.writeDerivation(&buf, true, store)
+		if err != nil {
+			return nil, err
+		}
+
+		h := sha256.New()
+
+		_, err = h.Write(buf.Bytes())
+		if err != nil {
+			return nil, err
+		}
+
+		partialDrvHash = hex.EncodeToString(h.Sum(nil))
+	}
+
+	// We populate a new map of outputPaths, which is what we return in the end of the function
+	outputPaths := make(map[string]string)
+
+	var err error
+	for outputName, o := range d.Outputs {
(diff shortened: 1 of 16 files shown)


pkg/derivation/output_path.go
```diff
@@ -0,0 +1,90 @@
...
+			return nil, err
+		}
+	}
+
+	return outputPaths, nil
+}
+
+// getSubstitutionHash is producing a hex-encoded hash of the current derivation.
+// It needs a store, as it does some substitution on the way.
+func (d *Derivation) getSubstitutionHash(store Store) (string, error) {
```
flokli, line 58: I think we want to move this method to the Derivation store.

Right now we recursively calculate this potentially multiple times.

This only needs to be calculated once for each derivation, so stores might populate some internal cache, either while calling `Put()`, or lazily on-first-access.

adisbladis: I really don't like this to be a function on the drv store implementation.
I think we need to reconsider this.

flokli: The Derivation store function only is a trampoline to `func (d *Derivation) OutputPaths(store Store)`.

All hash computation logic is still in `Derivation`, we just pipe the requests through the Derivation Store interface, so the calculation can be cached, and we can store the intermediate results to recuse less.

adisbladis (changes requested):

pkg/derivation/output_path.go
flokli: -> https://github.com/nix-community/go-nix/pull/58#discussion_r891231379

pkg/derivation/store.go
```diff
@@ -0,0 +1,27 @@
...
+type Store interface {
+	// Get retrieves a derivation by drv path.
+	// The second return argument specifies if the derivation could be found,
+	// similar to how acessing from a map works.
+	Get(string) (*Derivation, error)
+
+	// GetSubstitutionHash produces a hex-encoded hash of the current derivation.
+	// It recursively does this for all Input Derivations, so implementations might
+	// want to cache these results.
+	GetSubstitutionHash(string) (string, error)
```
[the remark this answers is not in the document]
flokli: cross-linking https://github.com/nix-community/go-nix/pull/58#discussion_r890444441

flokli:
@adisbladis I think you're right.

The hash calculation doesn't belong in the derivation store, and it doesn't belong in the derivation struct either.

We started to walk a painful path when we extended `writeDerivation` with specific hash-related arguments, and started to ping-pong references to the store around.

I think we should have a separate component, with access to a derivation store.

This component, when asked to calculate output hashes, will query the derivation store, and take care of all the hashing.

I looked at the derivation serializer again - the actual serialization behaviour (ordering, format, …) is exactly the same in all cases - it's just in hashing a context, the Derivation struct *values* need to look different.

This means, if we have something else driving the hashing, and crafting "temporary masked out Derivation structs", the individual `Derivation` doesn't need to know about or pass around a store, and the logic gets much simpler.

Such component can also then take care of caching intermediate hashes, without having to duplicate that code in each Derivation store.

Side note: the `Derivation.DrvHash()` function can still live in the `Derivation` struct - there's no different "variant" of the Derivation ATerm hashed into the drvpath, or knowledge about other derivations needed.

adisbladis:
Superseded by #63.

PR #61, opened 2022-06-07
pkg/derivation: use nixpath.Absolute to construct nix storepaths

flokli:
Make this super clear that we want forward slashes here.

While filepath.Join on Windows seemingly also does the right thing if
the first element already contains forward slashes, make it more clear
by using path.Join explicitly.


zimbatm:
Funny that it breaks the windows test :)

flokli:
I realized I was using `filepath.Join` in a place where I shouldn't have.

I moved `nixpath.Absolute` from https://github.com/nix-community/go-nix/pull/58 into here, added a bunch of tests, so the assumptions are clear, and started using `nixpath.Absolute` here. PTAL.

pkg/nixpath/nixpath_test.go
```diff
@@ -39,3 +41,27 @@ func TestNixPath(t *testing.T) {
...
+		assert.Equal(t, nixpath.StoreDir+"/"+"foo", s)
+	})
+	// test you can use this to exit nixpath.StoreDir
+	// Note path.Join does a path.Clean already, this is only
+	// written for additional clarity.
+	t.Run("leave storeDir", func(t *testing.T) {
+		s := nixpath.Absolute("..")
+		assert.Equal(t, path.Clean(path.Join(nixpath.StoreDir, "..")), s)
+		assert.False(t, strings.HasPrefix(s, nixpath.StoreDir),
+			"path shouldn't have the full storedir as prefix anymore (/nix)")
```
flokli, line 65: This is written a bit more clunky than it might need to be - however I want to support eventually building `go-nix` with another value of `nixpath.StoreDir` constant, and don't want to break the tests in that case.

PR #63, opened 2022-06-10
pkg/derivation: implement Output Path calculation (v3)

flokli:
More polished version than https://github.com/nix-community/go-nix/pull/58, adressing https://github.com/nix-community/go-nix/pull/58#issuecomment-1149016290.

I took another step back, and instead of bouncing around a `store` around between functions, went with a simple `map[string]string` for possible substitutions.

There's some more cleanup and restructuring, to make the functions cleaner and easier to understand.

I'll leave the `FSStore`/`HTTPStore`, and CLI additions to a follow-up PR.

### DIFF
--- a/pkg/derivation/encode.go
+++ b/pkg/derivation/encode.go
@@ -141,18 +141,16 @@ func (d *Derivation) writeDerivation(
 	if len(inputDrvReplacements) == 0 {
 		inputDerivations = d.InputDerivations
 	} else {
-		// this most certainly means someone is using this incorrectly
-		if len(inputDerivations) != len(d.InputDerivations) {
-			return fmt.Errorf("len(inputDrvReplacement) != len(d.InputDerivations)")
-		}
 		inputDerivations = make(map[string][]string, len(d.InputDerivations))
-		for drvPath, replacement := range inputDrvReplacements {
-			// get output names
-			outputNames, ok := d.InputDerivations[drvPath]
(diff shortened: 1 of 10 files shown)


adisbladis (approved):

pkg/derivation/encode.go
```diff
@@ -77,12 +78,27 @@ func encodeArray(opening byte, closing byte, quote bool, elems ...[]byte) []byte
...
+//   and `env[$outputName]` with empty strings
+//
+// - inputDrvReplacements (map[$drvPath]$replacement) can be provided.
+//   If set, it must contain all derivation path in d.InputDerivations[*]
+//   These will be replaced with their replacement value.
+//   As this will change map keys, and map keys need to be serialized alphabetically sorted,
+//   this will shuffle the order of values.
+//
+// This replacement/stripping is only used when calculating output hashes.
+// Set to false / nil in normal mode.
```
zimbatm, line 94: It seems like this is something we might want to hide. WriteDerivation would only get a io.Writer, but then call writeDerivation that takes the stripOutputs and inputDrvReplacements, with empty values

flokli: Indeed, we don't use these modes outside code in `package derivation` anymore.

These modes are only used in `getMaskedATermHash` (which is invoked by `CalculateOutputPaths`), and `CalculateDrvReplacement`.

PR #64, opened 2022-06-11
pkg/derivation/store: add BadgerStore

flokli:
This adds another `derivation.Store` implementation, using Badger under the hood.

Some of the generic validation and checking logic is moved into shared functions.

Badger provides some transactional guarantees, as well as persistence (if used in non-in-memory mode).

The `derivation.Store` was extended a bit - there's now also a `Has(context.Context, string) (bool, error)` function to check for the existence of something without retrieving it's value (which can be done faster in badger and other backends), as well as a `Close()` method.


pkg/derivation/store/badger.go
```diff
@@ -0,0 +1,200 @@
...
+// All derivations are stored in ATerm format, at `drv:$drvPath`.
+// The replacement string for a drv is stored at `replacement:$drvPath`.
+// The interface should be thread-safe.
+type BadgerStore struct {
+	db *badger.DB
+}
+
+// Put inserts a new Derivation into the Derivation Store.
+func (bs *BadgerStore) Put(ctx context.Context, drv *derivation.Derivation) (string, error) {
+	if err := validateDerivationInStore(ctx, drv, bs); err != nil {
```
flokli, line 50: This does trigger requests to `store.Has()` under the hood.

Right now, we don't support removing anything from a `derivation.Store`, but in the future, once we do, we might need to think about this, and potentially expand the transaction to this as well (and then maybe inline the code checking for these input derivations).

If we succeed all checks, and actually insert, we might also need to detect if something else removed any derivation this derivation (transitively) refers to.

Maybe that's also all too complicated, and `Put()` and `Delete()`/`GC()` should just block, I'm not sure - mostly leaving this here for future reference.

PR #65, opened 2022-06-12
test/testdata: document how fixtures are produced, fix json serialization

flokli:
Committing the original nix source files will allow reproducing test
fixtures both with Nix, as well as eventually our own implementation,
providing a nice closed loop.

This also fixes JSON serialization to match the output of `nix show-derivation`, and adds a test for it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Mark json files inside test/testdata as binary,
+# so git won't replace \n with \r\n on windows.
+test/testdata/*.json binary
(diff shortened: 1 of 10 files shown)


flokli:
@adisbladis do you still some more nix code for the other derivations used in `pkg/derivation/derivation_test.go`?

flokli:
I also fixed the JSON serialization to match what Nix outputs, and added a test to ensure it's the same.

Once we're able to (re)create more fixtures, we should be able to add more derivations.

PR #66, opened 2022-06-12
pkg/derivation/store: add HTTPStore, FSStore, CLI tooling

flokli:
~This depends on https://github.com/nix-community/go-nix/pull/64 and should be rebased once merged.~ Part of this was already in #58, but excluded from #63.

It introduces `pkg/derivation/store.{HTTPStore,FSStore}`, a `store.NewFromURI` method, and a `gonix drv --drv-store=uri-goes-here show /nix/store/drv-path.drv` command.

`{HTTPStore,FSStore}` don't implement any checking of output paths and referred derivations to exist, but I intend to provide this functionality via some "proxy store", that can be used to cache "remote" `derivation.Store` into a local one, and rely on the local store checks (left as a followup)

### DIFF
--- a/pkg/derivation/store/http.go
+++ b/pkg/derivation/store/http.go
@@ -1,0 +1,139 @@
+package store
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/nix-community/go-nix/pkg/derivation"
+)
+
+// HTTPStore implements derivation.Store.
+var _ derivation.Store = &HTTPStore{}
+
+// NewHTTPStore returns a HTTPStore with a given base URL.
+func NewHTTPStore(baseURL *url.URL) *HTTPStore {
+	return &HTTPStore{
+		Client:  &http.Client{},
+		BaseURL: baseURL,
+	}
+}
+
+// HTTPStore provides a store exposing all .drv files
+// directly hosted below the a HTTP path specified by baseURL
+// aka ${baseURl}/${base derivationPath}.
+// It doesn't do any output path validation and consistency checks,
+// meaning you usually want to wrap this in a validating store.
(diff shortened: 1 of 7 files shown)


flokli:
Rebased, as #64 was merged.

pkg/derivation/store/http.go
```diff
@@ -0,0 +1,139 @@
...
+	if resp.StatusCode == 404 {
+		return false, nil
+	}
+
+	// if we get back something in the 2xx range, this means
+	// the file exists
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return true, nil
+	}
+
```
adisbladis, line 131: This code should deal with redirects.

adisbladis (changes requested):

pkg/derivation/store/http.go
flokli: It already does! If you look further above, `hs.Client` is the default `http.Client{}`, so it reuses cached TCP connections, and also handles cookies and redirects.

From `go doc http.Client`:

```
    A Client is an HTTP client. Its zero value (DefaultClient) is a usable
    client that uses DefaultTransport.

    The Client's Transport typically has internal state (cached TCP
    connections), so Clients should be reused instead of created as needed.
    Clients are safe for concurrent use by multiple goroutines.

    A Client is higher-level than a RoundTripper (such as Transport) and
    additionally handles HTTP details such as cookies and redirects.

    When following redirects, the Client will forward all headers set on the
    initial Request except:
```

I opted to create a new `http.Client`, instead of using the default one, so the cookiejar and authentication data stays scoped to each of the stores. If the need arises, we can allow passing in another `http.Client` from the constructor

I think it however makes sense to explicitly add tests for this (probably also with a redirect example). I'll come up with a `httptest` server.

adisbladis: All good then!

PR #67, opened 2022-06-12
Fixes for various discrepancies in output hash calculation

adisbladis:


adisbladis:
I'm splitting this up into multiple PRs with test cases.

PR #70, opened 2022-06-12
pkg/derivation: Fix string quoting inside nested objects

adisbladis:


pkg/derivation/derivation_test.go
```diff
@@ -70,6 +71,29 @@ func TestParser(t *testing.T) {
...
+	t.Run("NestedJson", func(t *testing.T) {
+		derivationFile, err := os.Open("../../test/testdata/292w8yzv5nn7nhdpxcs8b7vby2p27s09-nested-json.drv")
+		if err != nil {
+			panic(err)
+		}
+
+		drv, err := derivation.ReadDerivation(derivationFile)
+		if err != nil {
+			panic(err)
+		}
```
flokli, line 84:
```suggestion
                assert.NoError(t, err, "reading a derivation with nested JSON shouldn't panic")
```

pkg/derivation/derivation_test.go
```diff
@@ -70,6 +71,29 @@ func TestParser(t *testing.T) {
...
+
+		nested := &struct {
+			Hello string
+		}{}
+
+		err = json.Unmarshal([]byte(drv.Env["json"]), &nested)
+		if err != nil {
+			panic(err)
+		}
```
flokli, line 91:
```suggestion
		err = json.Unmarshal([]byte(drv.Env["json"]), &nested)
		assert.NoError(t, err, "It should still be possible to parse the JSON afterwards")
```

pkg/derivation/parser.go
```diff
@@ -288,5 +289,10 @@ func arrayEach(value []byte, callback func(value []byte, index int) error) error
 }
 
 func unquote(b []byte) string {
-	return string(b[1 : len(b)-1])
+	s, err := strconv.Unquote(string(b))
+	if err != nil {
+		panic(err)
+	}
```
flokli, line 295: Can we wrap the error and extend the signature to pass it along, rather than panic-ing?

PR #71, opened 2022-06-12
pkg/derivation: Fix assumptions around name for structured attrs

adisbladis:
This PR in it's current form is more to demonstrate the issue than to propose an actual fix.

Depends on https://github.com/nix-community/go-nix/pull/70.

### DIFF
--- a/pkg/derivation/derivation.go
+++ b/pkg/derivation/derivation.go
@@ -14,6 +14,11 @@ import (
 // that the `nix show-derivation /path/to.drv` is using,
 // even though this might change in the future.
 type Derivation struct {
+	// Structured don't have the env name right in the regular spot but in the nested JSON object.
+	// This is an internal variable only used for structured attrs derivations, which can currently only be created
+	// from an existing drv file.
+	name string
+
 	// Outputs are always lexicographically sorted by their name (key in this map)
 	Outputs map[string]*Output `json:"outputs"`
 
@@ -122,6 +127,11 @@ func (d *Derivation) Validate() error {
 		if k == "name" {
 			hasNameEnv = true
 		}
+
+		// Structured attrs
+		if k == "__json" {
+			hasNameEnv = d.name != ""
+		}
 	}
 
 	if !hasNameEnv {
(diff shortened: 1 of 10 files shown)


pkg/derivation/derivation.go
```diff
@@ -122,6 +123,26 @@ func (d *Derivation) Validate() error {
 		if k == "name" {
 			hasNameEnv = true
 		}
+
+		// Structured don't have the env name right in the regular spot but in the nested JSON object.
+		// I don't particularly like to parse this object as a validation step so maybe we can just the name check
+		// if a derivation uses structured attrs.
```
flokli, line 129:
```suggestion
		// I don't particularly like to parse this object as a validation step so maybe 
		// we can only do the name check if a derivation uses structured attrs.
```

pkg/derivation/derivation.go
```diff
@@ -122,6 +123,26 @@ func (d *Derivation) Validate() error {
...
+			type structuredAttrs struct {
+				Name string
+			}
+
+			attrs := &structuredAttrs{}
+
+			err := json.Unmarshal([]byte(d.Env[k]), &attrs)
+			if err != nil {
+				return err
+			}
```
flokli, line 140: Wrap the error with a useful error message what went wrong here.

pkg/derivation/drv_path.go
```diff
@@ -54,10 +55,26 @@ func (d *Derivation) DrvPath() (string, error) {
...
+		type structuredAttrs struct {
+			Name string
+		}
+
+		attrs := &structuredAttrs{}
+
+		err := json.Unmarshal([]byte(structuredJson), &attrs)
+		if err != nil {
+			return "", err
+		}
```
flokli, line 69: Wrap the error with a useful error message what went wrong.

pkg/derivation/hashes.go
```diff
@@ -36,9 +37,27 @@ func (d *Derivation) getMaskedATermHash(inputDrvReplacements map[string]string)
...
+		type structuredAttrs struct {
+			Name string
+		}
+
+		attrs := &structuredAttrs{}
+
+		err := json.Unmarshal([]byte(structuredJson), &attrs)
+		if err != nil {
+			return nil, err
+		}
```
flokli, line 52: same

flokli:
It's terrible we need to do this, but I think we should - and definitely need to do while calculating the derivation path.

Maybe we can extract the "get name of a derivation" part into a private function for now?

flokli:
@adisbladis are you fine with merging this for now?

We can rework this later if we come up with a better way to convey this, but this should fix some of the hash calculations for now.

adisbladis:
Yes this is fine for now.

PR #74, opened 2022-06-14
pkg/derivation: Reduce the number of allocations by using unsafe

adisbladis:
These are some quick performance wins.

Ideally we'd rewrite the whole thing to write directly to the writer
and avoid allocations entirely.


pkg/derivation/encode.go
```diff
@@ -214,7 +212,7 @@ func (d *Derivation) writeDerivation(
 		}
 	}
 
-	_, err := writer.Write([]byte("Derive"))
+	_, err := writer.Write(derivationPrefix)
```
flokli, line 215: That's unrelated. Separate commit?

pkg/derivation/encode.go
```diff
@@ -17,30 +19,26 @@ var stringEscaper = strings.NewReplacer(
 	"\"", "\\\"",
 )
 
+func unsafeGetBytes(s string) []byte {
+	return (*[0x7fff0000]byte)(unsafe.Pointer(
```
flokli, line 23: what does the `0x7fff0000` magic number do? Does this also work properly on 32 bit architectures, or these with different endianess?

adisbladis: It's the max length of the byte slice, the version you commented on would have been broken with 32 bit, but the current version isn't.
Not sure about endianess here.

Ideally we'd bump the version in go.mod to 1.17 and use `unsafe.Slice`.

flokli: Do it :-) We're only testing Go 1.16 and 1.17 right now - might want to use 1.17 and 1.18 there, and bump the requirements to 1.17.

adisbladis: #75

pkg/derivation/encode.go
adisbladis: The motivation was to remove an allocation, it's related.

flokli:
With #75 merged, can you update this?

PR #76, opened 2022-06-15
pkg/derivation: Reduce allocations in writeDerivation

adisbladis:
Instead of allocating a bunch of arrays in which we encode the output
we can write the output directly to the writer.

This sacrifices readability a bit and I'm not too happy about the
excessive error handling required but I think it's worth it given
the quite drastic performance improvement.

Using the following program for testing:
``` go
package main

import (
	"fmt"
	"os"
	"testing"
	"time"

)

type nullWriter struct{}

func (n *nullWriter) Write(b []byte) (int, error) {
	return len(b), nil
}

func main() {

	n := 50

	if err != nil {
		panic(err)
	}

	drv, err := derivation.ReadDerivation(derivationFile)
	if err != nil {
		panic(err)
	}

	writer := &nullWriter{}

	{
		start := time.Now()

		for i := 0; i < n; i++ {
			err = drv.WriteDerivation(writer)
			if err != nil {
				panic(err)
			}
		}

	}

		err = drv.WriteDerivation(writer)
		if err != nil {
			panic(err)
		}
	})))

}
```

Yields the following performance numbers:
- Before
```
3.024143ms
Allocs: 15508
```

- After
```
1.279567ms
Allocs: 86
```

Note that I picked `texlive.combined.scheme-full` which is pretty much
the largest derivation file imaginable that is realistic and found out in
the wild.
The number of allocations is static and doesn't change when writing
another smaller derivation.

### DIFF
--- a/pkg/derivation/encode.go
+++ b/pkg/derivation/encode.go
@@ -1,7 +1,6 @@
 package derivation
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"reflect"
@@ -19,6 +18,16 @@ var stringEscaper = strings.NewReplacer(
 	"\"", "\\\"",
 )
 
+// nolint:gochecknoglobals
+var (
+	comma        = []byte{','}
+	parenOpen    = []byte{'('}
+	parenClose   = []byte{')'}
+	bracketOpen  = []byte{'['}
+	bracketClose = []byte{']'}
+	quoteC       = []byte{'"'}
+)
+
 func unsafeGetBytes(s string) []byte {
 	return unsafe.Slice(
 		(*byte)(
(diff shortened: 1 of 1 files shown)


pkg/derivation/encode.go
```diff
@@ -19,6 +18,16 @@ var stringEscaper = strings.NewReplacer(
 	"\"", "\\\"",
 )
 
+// nolint:gochecknoglobals
+var (
+	comma        = []byte{','}
```
zimbatm, line 23:
```suggestion
const (
	comma        = `,`
```
I wonder if there are performance improvements the compiler can find. If it knows that the value is immutable. This would mean casting the character to []byte on each usage though.

zimbatm (approved):

flokli (approved):

PR #78, opened 2022-06-16
pkg/nixpath: Reduce allocations when validating

adisbladis:
This is a follow-up to https://github.com/nix-community/go-nix/pull/77.

This ensures we don't have to do costly allocations for every store path that we validate when validating a derivation.

The following test program:
``` go
package main

import (
	"bytes"
	"fmt"
	"io"
	"os"
	"testing"
	"time"

)

func main() {

	n := 50

	if err != nil {
		panic(err)
	}

	drvBytes, err := io.ReadAll(derivationFile)
	if err != nil {
		panic(err)
	}

	{
		start := time.Now()

		for i := 0; i < n; i++ {
			_, err := derivation.ReadDerivation(bytes.NewReader(drvBytes))
			if err != nil {
				panic(err)
			}
		}

	}

		_, err := derivation.ReadDerivation(bytes.NewReader(drvBytes))
		if err != nil {
			panic(err)
		}
	})))

}
```

Yields the following numbers:
- Before
```
14.045441ms
Allocs: 27304
```

- After
```
8.503212ms
Allocs: 11864
```

### DIFF
--- a/pkg/nixbase32/nixbase32.go
+++ b/pkg/nixbase32/nixbase32.go
@@ -8,6 +8,58 @@ import (
 // Alphabet contains the list of valid characters for the Nix base32 alphabet.
 const Alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
 
+func decodeString(s string, dst []byte) error {
+	var dstLen int
+	if dst != nil {
+		dstLen = len(dst)
+	} else {
+		dstLen = DecodedLen(len(s))
+	}
+
+	for n := 0; n < len(s); n++ {
+		c := s[len(s)-n-1]
+
+		digit := strings.IndexByte(Alphabet, c)
+		if digit == -1 {
+			return fmt.Errorf("character %v not in alphabet", c)
+		}
+
+		b := uint(n * 5)
+		i := b / 8
+		j := b % 8
+
+		// OR the main pattern
+		if dst != nil {
+			dst[i] |= byte(digit) << j
+		}
+
+		// calculate the "carry pattern"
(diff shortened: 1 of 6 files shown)


pkg/nixbase32/nixbase32.go
```diff
@@ -8,6 +8,42 @@ import (
 // Alphabet contains the list of valid characters for the Nix base32 alphabet.
 const Alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
 
+// ValidateString validates if a byte slice is valid nixbase32.
+func Validate(b []byte) error {
```
flokli, line 12: The comment still stays `ValidateString`, but it's accepting bytes. Maybe we should call them `Validate{String,Bytes}`?

pkg/nixbase32/nixbase32.go
```diff
@@ -8,6 +8,42 @@ import (
...
+		carry := byte(digit) >> (8 - j)
+
+		// if we're at the end of dst…
+		if i == uint(dstLen-1) {
+			// but have a nonzero carry, the encoding is invalid.
+			if carry != 0 {
+				return fmt.Errorf("invalid encoding")
+			}
+		}
+	}
```
flokli, line 42: This is pretty much copy-pasted from `DecodeString`. Can we move this into a private function and use it from both places?

adisbladis: Done.

pkg/nixbase32/nixbase32.go
adisbladis: Done.

PR #79, opened 2022-06-16
pkg/derivation: Use unsafe to cast from []byte to string

adisbladis:
This ensures no copying.

The following test program:
``` go
package main

import (
	"bytes"
	"fmt"
	"io"
	"os"
	"testing"
	"time"

)

func main() {

	n := 50

	if err != nil {
		panic(err)
	}

	drvBytes, err := io.ReadAll(derivationFile)
	if err != nil {
		panic(err)
	}

	{
		start := time.Now()

		for i := 0; i < n; i++ {
			_, err := derivation.ReadDerivation(bytes.NewReader(drvBytes))
			if err != nil {
				panic(err)
			}
		}

	}

		_, err := derivation.ReadDerivation(bytes.NewReader(drvBytes))
		if err != nil {
			panic(err)
		}
	})))

}
```

Yields these numbers:
- Before:
```
13.011372ms
Allocs: 27305
```

- After:
```
12.119679ms
Allocs: 19537
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
-# Mark json files inside test/testdata as binary,
+# Mark everything inside test/testdata as binary,
 # so git won't replace \n with \r\n on windows.
-test/testdata/*.json binary
+test/testdata/* binary
+test/testdata/*.go binary diff
--- a/pkg/derivation/derivation_test.go
+++ b/pkg/derivation/derivation_test.go
@@ -151,6 +151,10 @@ func TestEncoder(t *testing.T) {
 			Title:          "latin1",
 			DerivationFile: "x6p0hg79i3wg0kkv7699935f7rrj9jf3-latin1.drv",
 		},
+		{
+			Title:          "cp1252",
+			DerivationFile: "m1vfixn8iprlf0v9abmlrz7mjw1xj8kp-cp1252.drv",
+		},
 	}
 
 	t.Run("WriteDerivation", func(t *testing.T) {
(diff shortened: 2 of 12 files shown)


zimbatm:
Are you 100% sure the []byte always contains only UTF-8?

adisbladis:
> Are you 100% sure the []byte always contains UTF-8?

That's a good point. I think realistically it always does but I've added a few test cases for other expressions with different encodings.

These actually fail when applied on master but passes here, presumably because on master these are erroneously decoded as utf-8 but here we'll just treat the bytes as utf-8 regardless.

flokli:
This might be git playing tricks - we should probably mark everything inside `test/testdata` as binary, to ensure any (wrong) binary encoding gets preserved.

@adisbladis can you edit `/.gitattributes` and see if this changes things?

adisbladis:
> This might be git playing tricks - we should probably mark everything inside `test/testdata` as binary, to ensure any (wrong) binary encoding gets preserved.
> 
> @adisbladis can you edit `/.gitattributes` and see if this changes things?

Done.

.gitattributes
```diff
@@ -1,3 +1,4 @@
-# Mark json files inside test/testdata as binary,
+# Mark everything inside test/testdata as binary,
 # so git won't replace \n with \r\n on windows.
-test/testdata/*.json binary
+test/testdata/* binary
+test/testdata/*.go text diff
```
flokli, line 4: This means if we ever embed `build-fixtures.go` in a test fixture itself, it'll look different on windows and linux :thinking: 

Maybe we should use `binary diff` here, to have git keep the file as-is, but still use `diff` to show differences?

adisbladis: Sure thing.

PR #81, opened 2022-06-19
pkg/nixpath/chunker: add

flokli:
This provides two different implementations to chunk data.


pkg/nixpath/chunker/fastcdc.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func NewFastCDCChunker(r io.Reader) (Chunker, error) { // nolint:ireturn
+	fastcdc.NewChunkerOptions()
+	chunkerOpts := fastcdc.NewChunkerOptions()
+
+	// FUTUREWORK: Test with different chunk sizes
+	chunkerOpts.NormalSize = 64 * 2024
+	chunkerOpts.MinSize = chunkerOpts.NormalSize / 4
+	chunkerOpts.MaxSize = chunkerOpts.NormalSize * 4
```
flokli, line 18: I just used the values here I used in nix-casync initially. This probably should still be refined once we can ingest a bit of data.

pkg/nixpath/chunker/chunker_test.go
```diff
@@ -0,0 +1,74 @@
...
+				chunk, err := chunker.Chunker.Next()
+				if err != nil {
+					if errors.Is(err, io.EOF) {
+						break
+					}
+					assert.NoError(t, err, "no other error other than EOF is accepted")
+				}
+				// check the chunk itself looks sane
+				assert.True(t,
+					uint64(len(chunk.Data)) == chunk.Size,
```
zimbatm, line 55: how about having chunk.Size() that returns len(chunk.Data), so the data could never by unsynchronized?

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/alecthomas/kong v0.5.0
 	github.com/dgraph-io/badger/v3 v3.2103.2
 	github.com/google/go-cmp v0.5.5
+	github.com/poolpOrg/go-fastcdc v0.0.0-20211130135149-aa8a1e8a10db
```
zimbatm, line 9: will you be able to re-use that and be casync-compatible or is that a new system altogether?

pkg/nixpath/chunker/chunker_test.go
flokli: I initially picked the interface provided by fastcdc, but you're right, it's redundant. I'll apply the suggestion.

go.mod
flokli: It's a new chunking mechanism, with a much smaller interface. You pass in a reader to some data and can use some iterator interface to get chunks, while reading through the data.

nix-casync used `desync`, which provided a lot of functionality that we didn't use (`.caidx`). Also, the way it was designed required us to first write the data to be chunked to a (temporary) file.

The chunking method used to chunk up data shouldn't matter when it comes to substitution. However, using the same chunking method with similar parameters should yield more block reuse.

pkg/nixpath/chunker/chunker_test.go
flokli: Updated.

flokli:
I marked this to a draft. I'm not entirely sure it belongs in `pkg/nixpath`, and if so, how to structure other things in the package.

Also, right now, `go-nix` mostly contains (re-)implementations of some common concepts found in Nix, and things like chunking mechanisms are not part of that. This might change once we have come up with a (already in-the-works) new remote store protocol.

flokli:
Closing in favor of https://github.com/nix-community/go-nix/pull/86.

PR #83, opened 2022-06-24
pkg/nar: Add benchmarks

adisbladis:


pkg/nar/util_test.go
```diff
@@ -8,43 +8,54 @@ import (
...
-			path2:    "/cmd/structlayout-optimize",
-			expected: true,
-		},
-		{
-			path1:    "/cmd/structlayout-optimize",
-			path2:    "/cmd/structlayout-ao/main.go",
-			expected: false,
-		},
-	}
+// nolint:gochecknoglobals
```
zimbatm, line 11: are there cases where this lint makes sense? It might be easier to set this globally otherwise.

zimbatm (approved):

pkg/nar/util_test.go
adisbladis: I think as a rule it makes sense to avoid global variables and we shouldn't introduce new ones without thinking twice, a linter rule is a good barrier I think.

PR #84, opened 2022-06-25
pkg/build: Add build package

adisbladis:
This adds a `pkg/build` package that implements building derivations.

It defines an interface for builds in general and implements a Linux builder using OCI for sandboxing.

TODO:
- [ ] `Store` is a stub interface, this should be broken out into a separate PR dealing with `Store` only.
- [ ] Figure out a `Store` interface for adding builds to store.
- [ ] Build tests
- [ ] Run as non-root inside container


pkg/build/linux/oci.go
```diff
@@ -0,0 +1,489 @@
...
+			return nil, err
+		}
+	}
+
+	containerUUID, err := uuid.NewRandom()
+	if err != nil {
+		return nil, fmt.Errorf("error creating container uuid: %w", err)
+	}
+
+	build.cmd = exec.CommandContext(ctx, ociRuntime, "run", containerUUID.String()) // nolint:gosec
```
zimbatm, line 444: Is that the OCI interface?

PR #87, opened 2022-07-13
pkg/nar: drop TypeUnknown

flokli:
Simply return an error when stumbling over something unknown during
DumpPath.

### DIFF
--- a/pkg/nar/dump.go
+++ b/pkg/nar/dump.go
@@ -59,7 +59,7 @@ func dumpPath(nw *Writer, path string, subpath string, filter SourceFilterFunc) 
 	} else if fi.Mode().IsRegular() {
 		nodeType = TypeRegular
 	} else {
-		nodeType = TypeUnknown
+		return fmt.Errorf("unknown type for %v", p)
 	}
 
 	if filter != nil && !filter(p, nodeType) {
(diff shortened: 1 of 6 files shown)


pkg/nar/dump.go
```diff
@@ -59,7 +59,7 @@ func dumpPath(nw *Writer, path string, subpath string, filter SourceFilterFunc)
 	} else if fi.Mode().IsRegular() {
 		nodeType = TypeRegular
 	} else {
-		nodeType = TypeUnknown
+		return fmt.Errorf("unknown type for %v", p)
 	}
 
 	if filter != nil && !filter(p, nodeType) {
```
flokli, line 65: @adisbladis does Nix' `builtins.filterSource` know an `unknown` here? Could we pass this as strings?

I assume we want to allow users to filter out unknown stuff, and fail further down if it's still in the list of things to add to the NAR…

flokli: Looks like Nix fails ungracefully:

```
❯ nix-build nix/default.nix -A a
error (ignored): error: end of string reached
error: file '/some/path/a/f' has an unsupported type
(use '--show-trace' to show detailed location information)
```

(this was me using the following derivation, where `../a/f` is a FIFO:

```
  a = pkgs.stdenv.mkDerivation {
    name = "blub";
    src = builtins.filterSource
      (path: type: type != "directory" || baseNameOf path != ".svn")
      ../a;
  };
```

PR #90, opened 2022-08-19
pkg/hash: use go-multihash, provide writer

flokli:
Instead of using our own `HashTypeSha256`, `HashTypeSha512` constants, use
`go-multihash` and its identifiers. Also use the `GetHasher` function to get
a hasher for it.

Expose a `Digest()` function, that either returns the parsed Digest, or
the state of the internal hash function.

Expose a `Multihash()` function, returning the digest in multihash
encoding.

Rename `Hash.String()` to `Hash.NixString()`, to be able to distinct it from
other string representations, like SRI, or hex-encoded multihash.

Have `hash.Hash` implement `io.Writer`. Also keep track on the total number
of BytesWritten(), which will become handy for NAR file synthesis.


pkg/hash/hash.go
```diff
@@ -3,100 +3,32 @@
...
-}
-
-// MustParseNixBase32 returns a new Hash struct, by parsing a hashtype:nixbase32 string, or panics on error.
-func MustParseNixBase32(s string) *Hash {
-	h, err := ParseNixBase32(s)
-	if err != nil {
-		panic(err)
-	}
+	// Hash only populated if we construct the hash on our own.
+	hash hash.Hash
```
flokli, line 18: Should this be called `hasher` instead?

pkg/hash/pool.go
```diff
@@ -0,0 +1,32 @@
+package hash
+
+import (
+	"fmt"
+	"sync"
+)
+
+// NewPool returns a sync.Pool of a Hasher with the given hashName
+// It creates one hash to check the hash name is supported
+// (which is then put in the pool), to avoid panic()'ing later.
```
flokli, line 10:
```suggestion
// NewPool returns a sync.Pool of a Hash with the given hashType
// It creates one Instance to check the hash type is supported
// (which is then put in the pool), to avoid panic()'ing later,
// as the sync.Pool New() function doesn't allow errors.
```

zimbatm (approved):
looks legit, I don't have a strong opinion here

pkg/hash/hash.go
adisbladis: The interface is called `Hash` so I think it's fine.

pkg/hash/encode.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+
+	"github.com/multiformats/go-multihash"
+	mh "github.com/multiformats/go-multihash/core"
+	"github.com/nix-community/go-nix/pkg/nixbase32"
+)
+
+//nolint:gochecknoglobals
```
adisbladis, line 11: I'd inline this function, something like:
``` go
var hashtypeToNixHashString = func() map[int]string {
	m := make(map[int]string)

	m[mh.SHA1] = "sha1"
	m[mh.SHA2_256] = "sha256"
	m[mh.SHA2_512] = "sha512"

	return m
}()
```

adisbladis (approved):

PR #92, opened 2022-09-04
nar/narinfo: add fingerprint calculation

zimbatm:
This is used as a source to generate and verify signatures.

Based on https://github.com/nixos/nix/blob/0c7f213c87c11c6c0f6b2fa07ba96548f6ea85ad/src/libstore/store-api.cc#L1212-L1222


pkg/nar/narinfo/narinfo_test.go
```diff
@@ -107,6 +107,30 @@ func TestNarInfo(t *testing.T) {
...
+URL: nar/05ra3y72i3qjri7xskf9qj8kb29r6naqy1sqpbs3azi3xcigmj56.nar.xz
+Compression: xz
+FileHash: sha256:05ra3y72i3qjri7xskf9qj8kb29r6naqy1sqpbs3azi3xcigmj56
+FileSize: 68852
+NarHash: sha256:1b4sb93wp679q4zx9k1ignby1yna3z7c4c2ri3wphylbc2dwsys0
+NarSize: 196040
+References: 0jqd0rlxzra1rs38rdxl43yh6rxchgc6-curl-7.82.0 6w8g7njm4mck5dmjxws0z1xnrxvl81xa-glibc-2.34-115 j5jxw3iy7bbz4a57fh9g2xm2gxmyal8h-zlib-1.2.12 yxvjs9drzsphm9pcf42a4byzj1kb9m7k-openssl-1.1.1n
+Deriver: 5rwxzi7pal3qhpsyfc16gzkh939q1np6-curl-7.82.0.drv
+Sig: cache.nixos.org-1:TsTTb3WGTZKphvYdBHXwo6weVILmTytUjLB+vcX89fOjjRicCHmKA4RCPMVLkj6TMJ4GMX3HPVWRdD1hkeKZBQ==
+Sig: test1:519iiVLx/c4Rdt5DNt6Y2Jm6hcWE9+XY69ygiWSZCNGVcmOcyL64uVAJ3cV8vaTusIZdbTnYo9Y7vDNeTmmMBQ==`))
```
flokli, line 121: Can you move the narinfo contents up to a strNarinfo… constant, like the previous ones, also assert fingerprints from the narinfo files above? (and maybe add a version without references to make sure we get it right there, too).

pkg/nar/narinfo/types.go
```diff
@@ -88,6 +89,31 @@ func (n *NarInfo) String() string {
 	return buf.String()
 }
 
+// StoreDir derives in which prefix the store is located, based on the narinfo
+// StorePath field.
+func (n NarInfo) StoreDir() string {
+	i := strings.LastIndex(n.StorePath, "/")
+	return n.StorePath[:i]
+}
```
flokli, line 97: We currently have a `nixpath.StoreDir` constant (which currently is only configurable by compiling with a different value). Parsing .narinfo files with other store prefixes is not supported. I'd probably use `nixpath.StoreDir` here as well - once/if we make this configurable, it'll be easier to spot the usages, and having this method might provide confusion, and cause people to assume we already support this.

pkg/nar/narinfo/types.go
```diff
@@ -88,6 +89,31 @@ func (n *NarInfo) String() string {
...
+
+// Fingerprint is the digest that will be used with a private key to generate
+// one of the signatures.
+func (n NarInfo) Fingerprint() string {
+	f := "1;" +
+		n.StorePath + ";" +
+		n.NarHash.NixString() + ";" +
+		fmt.Sprintf("%d", n.NarSize) + ";"
```
adisbladis, line 105: Instead of `fmt.Sprintf` use `strconv.Itoa`, it allocates less and has better performance.

adisbladis (changes requested):

flokli (approved):

PR #94, opened 2022-09-06
pkg/narinfo: move from pkg/nar/narinfo

flokli:
There's not much of a reason to put this inside the nar package.
These files can be useful even without a .nar file around.


zimbatm (approved):
sounds good

PR #95, opened 2022-09-07
pkg/exp/store: init

flokli:


stv0g:
Would it make sense to add some parts of the libstore API here is as well?

See: https://github.com/numtide/nar-serve/tree/main/libstore

flokli:
This was mostly an experiment, and is currently a bit dead. The NAR model doesn't allow granular access, so something like the tvix store model would be more suitable for something like nar-serve.

zimbatm:
Now that tvix lives in a different repo, it would make sense to have libstore as part of this repo.

flokli:
> Now that tvix lives in a different repo, it would make sense to have libstore as part of this repo.

Thinking about the repository structure is something still on my to-do list. Feel free to put stuff in pkg/exp, to signal it's experimental and the structure might change.

flokli:
Just took a closer look at this PR. It contains an earlier version of what we now have in tvix-castore (for which there's also golang bindings at `code.tvl.fyi/tvix/castore-go` and `code.tvl.fyi/tvix/store-go`). The approach here is discontinued, so the PR can be closed.

PR #96, opened 2022-11-04
go-nix drv show: match behavior of `nix show-derivation`

amjoseph-nixpkgs:
CC: @flokli

Hopefully this isn't too awful; this is my first time writing Go, so take it easy on me...

This PR gets `go-nix drv show` to match the behavior of `nix show-derivation` in two cases where it didnt: escapification of HTML-escapables (`>`, `<`, etc) and wrapping the json with a top-level map whose key is the drvName.  It also adds a test covering HTML-escapables since none of the existing tests covered that case.

Looking back on this, matching cppnix's escaping behavior might be a WONTFIX.  No downstream json consumer should depend on whether or not characters are escaped  other than `\"`).  I can drop that patch from the PR if you prefer.

### DIFF
--- a/cmd/gonix/main_test.go
+++ b/cmd/gonix/main_test.go
@@ -21,10 +21,10 @@ func TestDrvShow(t *testing.T) {
 		oldOut := os.Stdout
 		os.Stdout = pw
 		var buf bytes.Buffer
-		err_pipe := make(chan error, 1)
+		errPipe := make(chan error, 1)
 		go func() {
 			_, err := io.Copy(&buf, pr)
-			err_pipe <- err
+			errPipe <- err
 		}()
 		main()
 		os.Stdout = oldOut
@@ -32,7 +32,7 @@ func TestDrvShow(t *testing.T) {
 		if err != nil {
 			panic(err)
 		}
-		err = <- err_pipe
+		err = <- errPipe
 		if err != nil {
 			panic(err)
 		}
--- a/cmd/gonix/main_test.go
+++ b/cmd/gonix/main_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestDrvShow tests the `go-nix drv show` command.
+func TestDrvShow(t *testing.T) {
+	drvs := []string {"../../test/testdata/7dsin4zpzi50wv7nia5nsi813nyrjgal-escape_html.drv"};
+	for _, drvBasename := range drvs {
+		os.Args = []string{"gonix", "drv", "show", ("../../test/testdata/" + drvBasename)}
+		pr, pw, err := os.Pipe()
+		if err != nil {
+			panic(err)
+		}
+		oldOut := os.Stdout
+		os.Stdout = pw
+		var buf bytes.Buffer
+		err_pipe := make(chan error, 1)
+		go func() {
+			_, err := io.Copy(&buf, pr)
+			err_pipe <- err
+		}()
+		main()
+		os.Stdout = oldOut
+		err = pw.Close()
+		if err != nil {
+			panic(err)
+		}
+		err = <- err_pipe
+		if err != nil {
+			panic(err)
+		}
+		// compare the output with the prerecorded json output
+		derivationJSONFile, err := os.Open(filepath.FromSlash("../../test/testdata/" + drvBasename + ".json"))
+		if err != nil {
+			panic(err)
+		}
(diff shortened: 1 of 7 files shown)


flokli:
I don't really have a strong opinion on whether to escape `<` , `>`  and possibly other html characters.

About the "showing the derivation path as a top-level key": I was always confused why Nix was doing that. Is there any code path that outputs multiple keys?

cmd/gonix/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"path/filepath"
+	"testing"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestDrvShow tests the `go-nix drv show` command.
+func TestDrvShow(t *testing.T) {
+	drvs := []string {"../../test/testdata/7dsin4zpzi50wv7nia5nsi813nyrjgal-escape_html.drv"};
+	for _, drvBasename := range drvs {
+		os.Args = []string{"gonix", "drv", "show", ("../../test/testdata/" + drvBasename)}
```
flokli, line 16: I'm not entirely sure we should invoke the binary here, or the handlers directly (`cmd/gonix/drv.ShowCmd.Run`).

I guess we currently have to, because we can't specify another output location other than `os.Stdout`?

amjoseph-nixpkgs:
> About the "showing the derivation path as a top-level key": I was always confused why Nix was doing that. Is there any code path that outputs multiple keys?

I don't think so.

But there are a lot of awesome tools (like [nix-tree](https://github.com/utdemir/nix-tree)!) that work by parsing the json output of cppnix... in order for those to work we have to match the structure of those json outputs, or at least provide an environment variable `EMULATE_QUESTIONABLE_CPPNIX_DECISIONS=1` that does.

amjoseph-nixpkgs:
@flokli, would you mind putting me on the list of "not a cryptominer" users so I don't have to wait for somebody to run CI manually?

I'm not sure if 2d8969a will fix the CI failure, and even if it does I'm not sure that's the right solution.  The `go test` scripts expect all of the `.drv` files in `test/testdata/` to also be in `/nix/store` *on each CI runner*.  I suspect that because "check test fixtures are up to date" is a separate CI job, it runs on one of the runners but not necessarily all of them (and probably gets rolled back anyways).

flokli:
> @flokli, would you mind putting me on the list of "not a cryptominer" users so I don't have to wait for somebody to run CI manually? I'm on that list for `nixos/nix` and `nixos/nixpkgs`.

Sorry, this is your "first-time contributer" bit in this repo. I don't have permissions to change this, you need to get your first PR in ;-)

flokli:
> I'm not sure if [2d8969a](https://github.com/nix-community/go-nix/commit/2d8969a21511224b729e2c6cfe247fcdbba0b8b3) will fix the CI failure, and even if it does I'm not sure that's the right solution. The `go test` scripts expect all of the `.drv` files in `test/testdata/` to also be in `/nix/store` _on each CI runner_.

You don't have to. `gonix drv show` has a `--drv-store` command line argument, and instead of using the default `/nix/store`, you can point it to the fixture folder.

amjoseph-nixpkgs:
> You don't have to. `gonix drv show` has a `--drv-store` command line argument, and instead of using the default `/nix/store`, you can point it to the fixture folder.

Ah, thanks.  I will update this when I get a chance.  May be a few days.

flokli:
No worries. I was also thinking, instead of trying to get `gonix drv show` to behave like nix, what about explicitly having a `... show-derivation` command? Maybe in a separate `cmd` package? Then it'll become clear this hosts entrypoints that should behave exactly like nix does.

flokli:
In the meantime, Nix' `nix show-derivation` seems to fail on non-UTF8 strings:

```
❯ nix-instantiate test/testdata/cp1252.nix
warning: you did not specify '--add-root'; the result might be removed by the garbage collector
/nix/store/m1vfixn8iprlf0v9abmlrz7mjw1xj8kp-cp1252.drv
❯ nix show-derivation /nix/store/m1vfixn8iprlf0v9abmlrz7mjw1xj8kp-cp1252.drv
warning: 'show-derivation' is a deprecated alias for 'derivation show'
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/m1vfixn8iprlf0v9abmlrz7mjw1xj8kp-cp1252.drv^*'
error: [json.exception.type_error.316] invalid UTF-8 byte at index 1: 0xC4
```

Maybe we can just fail outputting JSON too

flokli:
@amjoseph-nixpkgs considering the :+1:, do you want to update this PR with https://github.com/nix-community/go-nix/pull/96#issuecomment-1685088235 ?

PR #100, opened 2022-12-25
test/testdata: regenerate JSON files

flokli:
It looks like the keys in `nix show-derivation` output were not sorted (but stable), and some Nix version update changed that.

Regenerate the JSON files to prevent CI from failing.

### DIFF
--- a/cmd/gonix/drv/cmd.go
+++ b/cmd/gonix/drv/cmd.go
@@ -28,6 +28,9 @@ func (cmd *ShowCmd) Run(drvCmd *Cmd) error {
 		return err
 	}
 
+	// Keep in mind `nix show-derivation` started sorting all of the JSON alphabetically,
+	// while this still preserves the previous order of keys, as  encoding/json
+	// preserves struct element definition order when serializing.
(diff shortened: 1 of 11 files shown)


zimbatm (commented):
sounds good but the tests are failing

flokli:
Yes, it looks like the golang serializer is not spitting out ordered keys, even though the golang docs suggest it should?!

flokli:
I sent a workaround.

```
    encoding/json doesn't serialize struct keys in alphabetical, but struct key definition order.
    
    This means we can't compare the raw bytes produced from Nix in the
    fixtures directly, but instead need to parse both in and compare again.
```

Maybe there is a way to properly get `encoding/json` to sort, which would be preferrable?

flokli:
I added a comment to the cmd entrypoint, and will merge this for now. Ideally, we'd probably replace `encoding/json` with something that does sort /all/ keys, but the CLI output being bit-by-bit compatible with what Nix does was never the intention of this anyways - and the test now checks for the JSON being equal.

PR #107, opened 2023-07-09
Update CI and repository settings

zombiezen:
- Set minimum Go version to 1.19, the current earliest supported version upstream. This has the effect of trimming the go.sum file because of a change in how modules are downloaded.
- Increase the timeout on golangci-lint. I've been seeing it fail in my experimentation.
- ~~The fixture-building seems to be nondeterministic.~~ I'm setting gitattributes to show more files as text in the interest of providing better diffs when the fixture test fails.


zimbatm:
thanks! It seems like changing the testdata introduced errors in CI.

zombiezen:
Yup. I tried to fix it, but the derivation code is complex enough that I'm going to instead avoid upgrading Nix for the fixture tests.

zombiezen:
Alright, I'm fairly confident this should pass now. (I've run the fixture generation with Nix 2.14 locally.)

flokli:
Oh, I did not see this PR before doing https://github.com/nix-community/go-nix/pull/109 and https://github.com/nix-community/go-nix/pull/110. Can check which of your CI-related fixes should still be ported from here?

zombiezen:
Done. PTAL

flokli (approved):

flokli (changes requested):
Hmmh, the linter complains about some discouraged use of a random number generator. Can you fix this up?

zombiezen:
Done.

flokli:
Thanks!

Issue #10, opened 2021-02-26
adopt golang-standards/project-layout 


flokli:
Link: https://github.com/golang-standards/project-layout

zimbatm:
So throw everything under `/internal` as it's unstable?

flokli:
No, I think some of the code is valuable to be used in a library context, but we should come up with a consistent structure.

I logged this as an action item for now, didn't put many thoughts into *how* the structure should look like.

flokli:
With https://github.com/numtide/go-nix/pull/23 in place, it feels like a good time. I try to tackle this in the coming days.

zimbatm:
sounds good!

PR #108, opened 2023-07-09
Fill in the rest of the base32 functions

zombiezen:
- Add `Encode` and `Decode` functions.
- Add `Is` function.
- Remove `ValidateBytes` function. It was unused AFAICT.
- Remove testify dependency in tests.
- Fix package comment.


README.md
```diff
@@ -33,7 +33,7 @@ A parser for .ls files (providing an index for .nar files)
 ## `pkg/nar/narinfo`
 A parser and generator for `.narinfo` files.
 
-## `pkg/nixbase32`
+## `nixbase32`
```
flokli, line 36: Why is this moved out of `pkg/`?

The idea for moving it into `pkg/` was in https://github.com/nix-community/go-nix/issues/10.

Also, please make the part moving it and changing things in there two separate commits, so the changes in the package itself can be easier reviewed.

nixbase32/nixbase32.go
```diff
@@ -0,0 +1,134 @@
...
+*/
+package nixbase32
+
+import (
+	"fmt"
+	"strings"
+)
+
+// alphabet contains the list of valid characters for the Nix base32 alphabet.
+const alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
```
flokli, line 17: This was exported because it's used in the regex in the nixpath package.

nixbase32/nixbase32_test.go
```diff
@@ -0,0 +1,226 @@
+package nixbase32
```
flokli, line 1: This was a separate test package to do the "black box testing" as described in https://pkg.go.dev/testing.

I also wanted to prevent certain test-related helpers to appear in the package itself.

Is there a reason this is discouraged now?

pkg/nixpath/nixpath.go
```diff
@@ -18,9 +18,8 @@ const (
 var (
 	NameRe = regexp.MustCompile(`[a-zA-Z0-9+\-_?=][.a-zA-Z0-9+\-_?=]*`)
 	PathRe = regexp.MustCompile(fmt.Sprintf(
-		`^%v/([%v]{%d})-(%v)$`,
+		`^%v/([0123456789abcdfghijklmnpqrsvwxyz]{%d})-(%v)$`,
 		regexp.QuoteMeta(StoreDir),
-		nixbase32.Alphabet,
```
flokli, line 23: See me above comment about why I left this exported.

README.md
zombiezen: See this [blog post on layout](https://eli.thegreenplace.net/2019/simple-go-project-layout-with-modules/) for rationale. golang-standards/project-layout is somewhat notorious for self-proclaiming that their layout recommendation is standard without actually having buy-in from the community, sadly.

That said, makes sense to have that be a separate PR. Reverted the change for now.

pkg/nixpath/nixpath.go
zombiezen: Can you help me understand why this regexp is exported? It's not used in this package, nor is it used by any importers AFAICT.

In my experience, exporting a regexp isn't a very flexible API. If it's being used for validation, then it prevents using higher-level validation later on.

nixbase32/nixbase32.go
zombiezen: Understood, but that appears to be the only reason? As per the [Go proverb](https://go-proverbs.github.io/), "a little copying is better than a little dependency." I've responded more on the `nixpath` package comment to evaluate whether that regex is necessary.

nixbase32/nixbase32_test.go
zombiezen: It isn't discouraged. From a practical standpoint, I wanted to use the now-unexported `alphabet` constant in the tests, which is why I made the change.

pkg/nixpath/nixpath.go
zombiezen: nixpath _is_ exporting a regex as a symbol, but I understand that `nixbase32.Alphabet` is just exporting a string, which is more reasonable.

That said, for something like the router usage you link, I would expect users to match on a `.narinfo` suffix and then check for existence on objects. `base32.ValidateString` can be used as a quick check that's more precise than using the alphabet for a regex. That's why I'm leaning toward unexporting the `Alphabet` constant: there are better ways for users of the library to accomplish the same goals.

flokli: I think I'd prefer keeping nixbase32.Alphabet exported for such usecases.

nixbase32/nixbase32_test.go
flokli: So the only reason on why we now have to move the tests into the package is because alphabet got unexported. As explained in https://github.com/nix-community/go-nix/pull/108#discussion_r1271294514, I'd prefer if it'd stay exported, and the tests a separate test package.

README.md
flokli: Interesting read, thanks! I'm not opposed to move things around. Can you squash your changes though, so this gets easier to review on a commit-by-commit basis?

flokli:
Can you keep `nixbase32.Alphabet` exported, undo the move of `nixbase32_test` into `nixbase32` (that's not needed then anymore), and then either squash the commits, or make them more granular on the changes that are kept?

nixbase32/nixbase32_test.go
zombiezen: Done.

pkg/nixpath/nixpath.go
zombiezen: Done.

zombiezen:
Re-exported, undone, and squashed.

flokli:
I fixed the current CI issues we had in https://github.com/nix-community/go-nix/pull/109 and https://github.com/nix-community/go-nix/pull/110. Can you rebase this one on top of current `main`? CI should then hopefully be green :-)

zombiezen:
Done.

flokli:
Can you rebase on top of latest `main` branch?

pkg/nixbase32/nixbase32.go
```diff
@@ -1,73 +1,94 @@
...
 }
 
 // EncodedLen returns the length in bytes of the base32 encoding of an input
 // buffer of length n.
 func EncodedLen(n int) int {
-	if n == 0 {
-		return 0
-	}
-
-	return (n*8-1)/5 + 1
```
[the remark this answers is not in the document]
zimbatm: It's the same algorithm with less branching and less operations.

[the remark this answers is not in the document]
edef1c: The branch wasn't needed to begin with: we're working in signed arithmetic, so it comes out to `(n*8-1)/5 + 1 = -1/5+1 = 0`. I'm not sure why 135836bf135c7b6ca476268bfea3e979a1e3a34b added it.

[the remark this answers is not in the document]
edef1c: This seems like the right move. We originally just matched the somewhat awkward arithmetic that the Nix upstream impl uses.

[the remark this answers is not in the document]
flokli: Note to self: this pushes the addition into the paranthesis, before dividing by 5, and still works for 0, because 4/5 in go ints is 0.

PR #111, opened 2023-08-08
feat: add narinfo info command

a-h:
Useful for examining the store path of narinfo files created by the `nix copy` command.

ls *.narinfo | xargs -I {} gonix narinfo info {} | grep StorePath


flokli:
Doesn't the String() method just plainly returns the same contents back that the .narInfo file already contained?

a-h:
Oh dear, why did I not think to look in the files first?

PR #125, opened 2024-12-06
narinfo: Deriver is optional, Check must not fail if empty

kalbasit:
The deriver field is optional so the `Check()` function should not return an error if it's empty.

References:
- [nix-serve source code](https://github.com/edolstra/nix-serve/blob/2641d7f6372572f82f8701f38c8cd5aaed0188b4/nix-serve.psgi#L37).
- [narinfo rust crate](https://docs.rs/narinfo/latest/narinfo/struct.NarInfo.html#structfield.deriver)

ref https://github.com/kalbasit/ncps/issues/51

### DIFF
--- a/pkg/narinfo/check.go
+++ b/pkg/narinfo/check.go
@@ -26,9 +26,11 @@ func (n *NarInfo) Check() error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks!

PR #128, opened 2025-01-01
fix(pkg/narinfo): Treat unknown-deriver as no deriver

kalbasit:
Some cache servers return `Deriver: unknown-deriver` which breaks verification of a narinfo. Treat such deriver as if no deriver was given at all.

ref https://github.com/kalbasit/ncps/issues/171


flokli:
Is this a cachix thing, or does nix/hydra also upload NARInfo like this?

flokli:
Apparently it's a Nix thing - see https://github.com/NixOS/nix/commit/c60715e937e3773bbb8a114fc9b9c6577f8c5cb5.

Thanks for fixing this, and also for adding a test!

flokli:
https://cl.tvl.fyi/c/depot/+/12940 applies the same change to nix-compat too.

PR #129, opened 2025-03-27
 Add support for recursive search

fzakaria:
The Nix, Lix & Snix (twix) codebase both support a recursive call for the lookup of the input derivation replacement paths.

Augment hashes.go to include the ability to recursively lookup derivations and calculate their paths.

A demonstration of this is done in the test but it can also be done by looking up derivations on the /nix/store filesystem for instance.


pkg/derivation/hashes.go
```diff
@@ -141,6 +141,18 @@ func (d *Derivation) CalculateOutputPaths(inputDrvReplacements map[string]string
 // We solve this having calculateDrvReplacement accept a map of
 // /its/ replacements, instead of recursing.
```
flokli, line 142: The diff is a bit hard to read, but this docstring does explicitly say we expect a (immutabe) hashmap with the hashes already precomputed. Doing a call to `CalculateDrvReplacementRecursive` inside then seems wrong.

flokli (commented):
It's been a while since I touched this code, but I'd feel more comfortable if the code would be moved around the bit, so `CalculateDrvReplacement` does not call `CalculateDrvReplacementRecursive`.

fzakaria:
@flokli up to you on the change.
I modeled it similarly to what I saw in Snix & CppNix which need recursive calls to follow the input derivations to get the replacements. 

I thought that the original code can be replaced with the recursive variant with a fixed function provider was a good demonstration of the interface ,😅🫠
(Writing from phone)

